PR #81, opened 2026-04-14
fix(executor): replace production stubs with on-chain validation (#66)

### LINKED ISSUE #66: [E1/WS-1] Remove executor stubs and fail-fast on missing config
## Context

`cmd/executor/main.go:39-40` has hardcoded stub values that will cause silent failures in production:

- `ExecutorAddr: "0x0..."` — zero address default means bundles route to a non-existent contract
- `EthBalance: 0.5` — static stub bypasses real balance monitoring; risk manager never halts on low balance
- `ChainID: 1` — hardcoded; no cross-check against the actual chain the node is connected to

### WS-1.1 — Remove ExecutorAddr stub
- Remove the `"0x0..."` default from `cmd/executor/main.go:39`
- Require `AETHER_EXECUTOR_ADDRESS` env var **or** `config/executor.yaml` field
- Validate the address is non-zero and has deployed bytecode (`eth_getCode != 0x`)
- Fail-fast at startup with a clear error message if missing or invalid

### WS-1.2 — Replace EthBalance stub with live polling
- Remove `EthBalance: 0.5` from `cmd/executor/main.go:40`
- Add a background goroutine that calls `eth_getBalance(searcher_address)` on every new block
- Feed the result into the risk manager's `EthBalanceCheck` circuit breaker
- The risk manager already halts at <0.1 ETH — this wires it to real data

### WS-1.3 — Remove hardcoded ChainID
- Remove `ChainID: 1` from `cmd/executor/main.go:37`
- Call `eth_chainId` at startup
- Cross-check against `config/executor.yaml` expected chain ID
- Fail-fast on mismatch (prevents accidental testnet deployment with mainnet config)

## Files

- `cmd/executor/main.go` — primary changes
- `config/executor.yaml` (new) — executor address, expected chain ID
- `internal/risk/manager.go` — wire live ETH balance into existing circuit breaker

## Acceptance Criteria

- [ ] Starting without `AETHER_EXECUTOR_ADDRESS` or `config/executor.yaml` exits with error
- [ ] Starting with a zero address exits with error
- [ ] ETH balance updates every block in Prometheus (`aether_eth_balance` gauge)
- [ ] Chain ID mismatch at startup exits with error
- [ ] `go test ./cmd/executor/... -race` passes
- [ ] Grep confirms no remaining `"0x0"` or `0.5` stubs in `main.go`

## Epoch

E1 — Stop the Bleeding (prod correctness). Must ship before E2.

0xfandom:
## Summary

Closes #66. Removes the three hardcoded stubs in `cmd/executor/main.go:37-40` that silently faked production state, and wires each one to real on-chain data with fail-fast startup validation.

- **ExecutorAddr** `"0x0000…"` → loaded from new `config/executor.yaml` (with `AETHER_EXECUTOR_ADDRESS` env override), verified against `eth_getCode != 0x` at startup.
- **EthBalance** `0.5` → replaced with a lock-free `LiveBalance` refreshed every 30s by `balanceWatchLoop` and read by `processArb` on every inbound arb, so the risk manager's `<0.1 ETH halt` finally sees real data.
- **ChainID** `1` → `expected_chain_id` in `executor.yaml`, cross-checked against `eth_chainId` at startup.
- **`ETH_RPC_URL`** is now required — all three checks above need it.

## Files Changed

| File | Change |
|---|---|
| `config/executor.yaml` | **new** — `executor_address` (supports `${AETHER_EXECUTOR_ADDRESS}` expansion) and `expected_chain_id` |
| `internal/config/loader.go` | new `ExecutorFileConfig`, `LoadExecutorConfig`, `ValidateExecutorConfig` |
| `internal/config/loader_test.go` | 4 new tests: valid config, 6 invalid cases, env expansion, missing file |
| `cmd/executor/main.go` | drop stubs; load executor.yaml; require `ETH_RPC_URL`; verify chain ID and executor bytecode; wire live balance into arb stream |
| `cmd/executor/metrics.go` | new `LiveBalance` (atomic float64 via `math.Float64bits`); refactor `balanceWatchLoop` to write it; extract `fetchAndStoreBalance` for shared use at startup and in the loop |
| `cmd/executor/integration_test.go` | update call sites to pass `*LiveBalance` instead of `float64` |

## Acceptance Criteria

| Criterion | Status | Evidence |
|---|---|---|
| Starting without `AETHER_EXECUTOR_ADDRESS` or `config/executor.yaml` exits with error | ✅ | Smoke: `FATAL: executor config (...): executor_address must not be empty` (yaml uses `${AETHER_EXECUTOR_ADDRESS}`; empty expansion → reject) |
| Starting with a zero address exits with error | ✅ | Smoke: `FATAL: ... executor_address must not be the zero address` |
| ETH balance updates in Prometheus (`aether_eth_balance` gauge) | ✅ (see note below) | `fetchAndStoreBalance` writes `ethBalanceGauge.Set(ethVal)` on every tick; called at startup and every 30s by `balanceWatchLoop` |
| Chain ID mismatch at startup exits with error | ✅ | `main.go` calls `ethClient.ChainID(ctx)`, compares to `execCfg.ExpectedChainID`, `log.Fatalf("FATAL: chain-id mismatch — node reports %d, config expects %d", ...)` |
| `go test ./cmd/executor/... -race` passes | ✅ | `ok github.com/aether-arb/aether/cmd/executor 9.725s` |
| No remaining `"0x0"` or `0.5` stubs in `main.go` | ✅ | `grep -nE '0x0{40}\|EthBalance:\s*0\.5\|ChainID:\s*1' cmd/executor/main.go` → no matches |

### Note on "every block" phrasing in the issue

The issue's acceptance criterion says "ETH balance updates **every block**". This PR ships a 30-second timer rather than a strict per-block refresh. Reason: the Go executor has no WebSocket subscription today — `ethclient.SubscribeNewHead` requires `ws://`/`wss://`/IPC, and all existing RPC usage in `cmd/executor/` is request/response. Adding WS wiring is real work beyond "remove stubs."

**Two follow-up approaches for strict per-block** (not in this PR):

1. **WebSocket subscription** — require `ETH_RPC_URL` (or a new `ETH_WS_URL`) to be a WS endpoint; call `SubscribeNewHead` and refresh balance on each head.
2. **Piggyback on the arb stream** — every `ArbOpportunity` already carries `arb.BlockNumber`; refresh balance whenever the block advances. Zero new infrastructure.

Happy to ship whichever the reviewer prefers as a follow-up, or widen this PR if you'd rather do it now.

## Test plan

- [x] `go test ./... -race -count=1` — all packages green (`cmd/executor` 9.7s under race)
- [x] `cargo test --workspace --release` — 0 failed across all crates
- [x] `forge test` — 38 passed, 0 failed, 2 skipped
- [x] `go vet ./...`, `go build ./...` — clean
- [x] Fail-fast smoke tests:
  - [x] no `ETH_RPC_URL` → `FATAL: ETH_RPC_URL not set — required for chain-id / bytecode / balance checks`
  - [x] no `AETHER_EXECUTOR_ADDRESS` → `FATAL: ... executor_address must not be empty`
  - [x] zero address env override → `FATAL: ... must not be the zero address`
  - [x] unreachable RPC → `FATAL: eth_chainId failed: ...`
- [ ] Production dry-run against a mainnet RPC with a real deployed `AetherExecutor` — pending deploy

### DIFF
--- a/cmd/executor/main.go
+++ b/cmd/executor/main.go
@@ -233,7 +233,8 @@ func main() {
 	// and bytecode checks.
 	if txSigner != nil {
 		if err := fetchAndStoreBalance(ctx, ethClient, txSigner.Address(), liveBalance); err != nil {
-			log.Fatalf("FATAL: initial eth_getBalance failed: %v", redactRPCError(err, rpcURL))
+			log.Fatalf("FATAL: initial eth_getBalance(%s) failed: %v",
+				txSigner.Address().Hex(), redactRPCError(err, rpcURL))
 		}
 		wg.Add(1)
 		go func() {
--- a/cmd/executor/main.go
+++ b/cmd/executor/main.go
@@ -95,19 +95,14 @@ func main() {
 	// Executor on-chain parameters (contract address, expected chain ID) are
 	// required: the service refuses to start without them. This prevents the
 	// old fail-open behaviour where a zero-address stub silently routed
-	// bundles to nowhere.
+	// bundles to nowhere. Deployments inject the address via
+	// ${AETHER_EXECUTOR_ADDRESS} which executor.yaml expands at load time —
+	// ExpandEnv runs inside LoadExecutorConfig before validation, so no
+	// separate post-load override path is needed.
 	execPath := config.ConfigPath("executor.yaml")
 	execCfg, err := config.LoadExecutorConfig(execPath)
 	if err != nil {
 		log.Fatalf("FATAL: executor config (%s) missing or invalid: %v", execPath, err)
-	}
-	// AETHER_EXECUTOR_ADDRESS env var overrides the yaml value so deployments
-	// can inject the deployed address without editing the file in-tree.
-	if envAddr := os.Getenv("AETHER_EXECUTOR_ADDRESS"); envAddr != "" {
-		execCfg.ExecutorAddress = envAddr
-	}
-	if err := config.ValidateExecutorConfig(execCfg); err != nil {
-		log.Fatalf("FATAL: executor config invalid after env override: %v", err)
 	}
 	log.Printf("Config: executor_address=%s expected_chain_id=%d",
 		execCfg.ExecutorAddress, execCfg.ExpectedChainID)
@@ -119,10 +114,11 @@ func main() {
 		log.Fatalf("FATAL: ETH_RPC_URL not set — required for chain-id / bytecode / balance checks")
 	}
 	dialCtx, dialCancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer dialCancel()
 	ethClient, err := ethclient.DialContext(dialCtx, rpcURL)
-	dialCancel()
 	if err != nil {
-		log.Fatalf("FATAL: failed to connect to ETH_RPC_URL: %v", err)
+		log.Fatalf("FATAL: failed to connect to ETH_RPC_URL (%s): %v",
+			redactRPCURL(rpcURL), redactRPCError(err, rpcURL))
 	}
 	log.Printf("Connected to Ethereum node")
 
@@ -130,10 +126,10 @@ func main() {
 	// executor.yaml. A mismatch here typically means someone pointed a
 	// mainnet config at a testnet RPC (or vice versa) — refuse to start.
 	chainCtx, chainCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer chainCancel()
 	chainID, err := ethClient.ChainID(chainCtx)
-	chainCancel()
 	if err != nil {
-		log.Fatalf("FATAL: eth_chainId failed: %v", err)
+		log.Fatalf("FATAL: eth_chainId failed: %v", redactRPCError(err, rpcURL))
 	}
 	if chainID.Int64() != execCfg.ExpectedChainID {
 		log.Fatalf("FATAL: chain-id mismatch — node reports %d, config expects %d",
@@ -144,10 +140,11 @@ func main() {
 	// Verify the configured executor contract actually exists on-chain. A
 	// zero-bytecode result means we'd be sending bundles to a non-contract.
 	codeCtx, codeCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer codeCancel()
 	code, err := ethClient.CodeAt(codeCtx, common.HexToAddress(execCfg.ExecutorAddress), nil)
-	codeCancel()
 	if err != nil {
-		log.Fatalf("FATAL: eth_getCode(%s) failed: %v", execCfg.ExecutorAddress, err)
+		log.Fatalf("FATAL: eth_getCode(%s) failed: %v",
+			execCfg.ExecutorAddress, redactRPCError(err, rpcURL))
 	}
 	if len(code) == 0 {
 		log.Fatalf("FATAL: executor address %s has no bytecode on chain %d",
@@ -228,16 +225,20 @@ func main() {
 		gasOracle.UpdateLoop(ctx, 12*time.Second)
 	}()
 
-	// Start ETH balance watcher. Perform one synchronous fetch first so the
-	// first arb does not see a zero balance and get rejected incorrectly.
+	// Start ETH balance watcher. The initial fetch is synchronous and fatal
+	// on failure: LiveBalance starts at zero, and risk.PreflightCheck rejects
+	// anything below MinETHBalance, so a transient startup blip would
+	// silently kill every arb for up to 30s until balanceWatchLoop's first
+	// tick. This matches the fatal-on-startup pattern of the dial, chain-ID,
+	// and bytecode checks.
 	if txSigner != nil {
 		if err := fetchAndStoreBalance(ctx, ethClient, txSigner.Address(), liveBalance); err != nil {
-			log.Printf("WARNING: initial eth_getBalance failed: %v", err)
+			log.Fatalf("FATAL: initial eth_getBalance failed: %v", redactRPCError(err, rpcURL))
 		}
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			balanceWatchLoop(ctx, ethClient, txSigner.Address(), 30*time.Second, liveBalance)
+			balanceWatchLoop(ctx, ethClient, txSigner.Address(), 30*time.Second, liveBalance, rpcURL)
 		}()
 	}
 
--- a/cmd/executor/main.go
+++ b/cmd/executor/main.go
@@ -12,6 +12,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 
 	"github.com/aether-arb/aether/internal/config"
@@ -21,23 +22,20 @@ import (
 )
 
 // Config holds executor service configuration.
+// ChainID, ExecutorAddr, and the live ETH balance are no longer carried here —
+// they are resolved against the connected node at startup (see main) and the
+// balance is updated continuously by balanceWatchLoop.
 type Config struct {
 	GRPCAddress    string
 	BuilderConfigs []BuilderConfig
-	ChainID        int64
 	MaxGasGwei     float64
-	ExecutorAddr   string  // On-chain AetherExecutor contract address
-	EthBalance     float64 // Current ETH balance of searcher wallet (simulated)
 }
 
 func defaultConfig() Config {
 	return Config{
 		GRPCAddress:    "localhost:50051",
 		BuilderConfigs: defaultBuilderConfigs(),
-		ChainID:        1,
 		MaxGasGwei:     300.0,
-		ExecutorAddr:   "0x0000000000000000000000000000000000000000",
-		EthBalance:     0.5,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


Pablosinyores (changes requested):
Thanks for tackling this — the shape of the fix (config + env override + startup validation + live balance) is the right approach, and the new tests in `loader_test.go` are solid. Pulled the branch and ran through it: `go test ./... -race -count=1` clean (159 pass), `go vet` / `go build` clean, `LiveBalance` concurrency holds up under stress. Two things need to change before merge, plus some smaller bits.

### Blockers

**1. RPC URL leaks into error logs.** go-ethereum wraps HTTP failures as `Post "<full URL>": <cause>`, and we pass `%v` of the error straight to `log` in four spots:

- `cmd/executor/main.go:136` — `log.Fatalf("FATAL: eth_chainId failed: %v", err)`
- `cmd/executor/main.go:150` — `log.Fatalf("FATAL: eth_getCode(%s) failed: %v", ...)`
- `cmd/executor/main.go:235` — `log.Printf("WARNING: initial eth_getBalance failed: %v", err)`
- `cmd/executor/main.go:257` — `log.Printf("WARNING: eth_getBalance failed: %v", err)` inside `balanceWatchLoop`

Alchemy/QuickNode/Infura all embed the API key in the URL path, so those errors end up in stdout, journald, Loki and alert payloads. Please redact before logging — simplest approach is to split the URL at the last `/`, mask the tail, and pass the sanitized form to the formatter; apply it uniformly across dial / chain-id / code-at / balance-at sites.

**2. `ValidateExecutorConfig` accepts non-hex characters.** `internal/config/loader.go:264-280` checks `len(addr) == 42` and `addr[:2] == "0x"` but not hex-ness. Verified locally that `"0xZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ"` passes validation, and `common.HexToAddress` silently coerces it to the zero address. The runtime `eth_getCode` guard at `main.go:152` saves us end-to-end (zero address has no bytecode), but defense-in-depth is the whole premise of this PR. Swap in `common.IsHexAddress(addr)` alongside the length check, `strings.TrimSpace` the raw string for good measure, and extend `TestValidateExecutorConfig_Errors` with a non-hex case.

### Worth fixing before merge

**3. Dead override block.** `cmd/executor/main.go:106-108` re-applies `AETHER_EXECUTOR_ADDRESS` after `LoadExecutorConfig` returns — but `LoadExecutorConfig` already runs `ExpandEnv → Validate`, so with the in-tree yaml template (`executor_address: ${AETHER_EXECUTOR_ADDRESS}`) and the env var unset, validation fails at line 102 and the override is unreachable. The comment at 104-105 calls it an "override" but what actually fires is env-substitution-at-load. Either drop the post-load block (env expansion already does the work) or pull validation out of `LoadExecutorConfig` so the override runs first. Option A is simpler.

**4. Silent arb rejection window on startup.** `main.go:233-242` treats initial `eth_getBalance` failure as a WARNING and continues. `LiveBalance` starts at zero and `risk.PreflightCheck` rejects anything below `MinETHBalance=0.1`, so a transient blip on startup silently kills every arb for up to 30s while ops stares at a `risk_rejections_total` spike wondering why. Inconsistent with the chain-id / bytecode / dial paths, which are all fatal. Either make initial fetch fatal too, or add a short retry (3× @ 1s), or poll on a shorter interval until the first success lands.

**5. Test gaps in env-expansion.** `TestLoadExecutorConfig_EnvExpansion` only covers the happy path. Add `TestLoadExecutorConfig_EnvUnset` (yaml references `${TEST_UNSET_VAR}`, env not set, assert "must not be empty") and `TestLoadExecutorConfig_EnvIsZeroAddress` (env set to zero addr, assert zero-address error). While you're in there, add the non-hex case for #2.

### Nits

**6.** YAML loader doesn't use strict decoding — a typo like `expected_chainid:` silently leaves the field zero and trips the `>0` validator, giving a technically-correct but unhelpful error. `yaml.Decoder.KnownFields(true)` would reject unknown keys and point at the real mistake.

**7.** `main.go:121-126` uses `dialCancel()` imperatively rather than `defer dialCancel()` (same pattern at 132-134 and 146-148). Fine today, brittle if anyone adds a retry in between.

**8.** `config/executor.yaml` comment says "override" — if you drop #3 it should say "env-var substitution".

### Merge conflict diagnosis

The `CONFLICTING` flag is only against `cmd/executor/main.go`, caused by `4d39159` on main (the error-surfacing commit that touched the same startup block). The other five files merge clean. Resolution is straightforward — keep this PR's "fatal on missing ETH_RPC_URL + always-non-nil ethClient" structure and fold in the error-surfacing improvements from `4d39159`.

### Verified good

- `LiveBalance` concurrency — `math.Float64bits` + `atomic.Uint64` is the canonical way, race detector clean under 8×8×10k stress.
- All three stubs removed from `main.go` (grep clean).
- chain-id check runs before `eth_getCode` (lines 133 vs 147), bytecode non-empty guard at 152-155 correctly covers the `"0x"` reply.
- `liveBalance.Get()` is read on every `processArb` call (line 421), no caching between arbs.

Happy to re-review once (1) and (2) land.

0xfandom:
Thanks for the thorough review. All items addressed in `5147e87`, rebased onto `main` (conflict resolved by folding `4d39159`'s `switch/case` error-surfacing into this PR's `liveBalance.Get()` call site).

### Blockers

**1. RPC URL redaction.** Added `redactRPCURL` / `redactRPCError` helpers in a new `cmd/executor/rpc_redact.go`. Masks everything past the scheme+host (Alchemy / Infura put the key as the last path segment, QuickNode as an intermediate segment — last-`/` masking leaks on the QuickNode shape, so host-only is the only form that's safe across providers). Applied at all four sites you called out plus the dial error (same go-ethereum `Post "<url>"` wrap path). Unit-tested end-to-end: nil passthrough, URL-less errors passthrough, wrapped `Post` errors redacted, and a parametric check that the Alchemy/Infura/QuickNode API keys never survive.

**2. `ValidateExecutorConfig` non-hex guard.** Now `strings.TrimSpace` the raw value, require the `0x` prefix explicitly (because `common.IsHexAddress` accepts bare 40-char hex too), and then `common.IsHexAddress(addr)`. Added `"non-hex chars"` and `"whitespace-only address"` cases to `TestValidateExecutorConfig_Errors`.

### Worth fixing

**3. Dead override block.** Dropped the post-load `AETHER_EXECUTOR_ADDRESS` reassignment. `LoadExecutorConfig` already runs `ExpandEnv → Validate`, so the in-tree yaml's `${AETHER_EXECUTOR_ADDRESS}` placeholder is all the override path we need. Updated `config/executor.yaml`'s comment to say "substitution at load time" instead of "override" to match reality.

**4. Silent arb rejection window.** Initial `eth_getBalance` is now `log.Fatalf` — consistent with dial / chain-id / bytecode. Chose the fatal path over retries so operators get an immediate signal rather than a 30s window of "why is everything rejected."

**5. Env-expansion test gaps.** Added `TestLoadExecutorConfig_EnvUnset` (expands to empty → rejected with the empty-address error), `TestLoadExecutorConfig_EnvIsZeroAddress` (expands to `0x000…` → rejected with the zero-address error), and the non-hex case lives in `TestValidateExecutorConfig_Errors` alongside the existing invariants.

### Nits

**6. Strict YAML.** Introduced `decodeStrictYAML(data, out)` in `loader.go` that wraps `yaml.NewDecoder + KnownFields(true)` and wired it into `LoadExecutorConfig`. New `TestLoadExecutorConfig_RejectsUnknownKey` covers `expected_chainid:` — now surfaces as a decoder error naming the offending field instead of the misleading `> 0` message. Kept the other loaders on the permissive path to avoid unrelated churn.

**7. `defer dialCancel`.** All three context cancels (`dial`, `chain`, `code`) are now `defer`. The contexts live for the duration of `main`, which is fine given they were already scoped to startup-only work.

**8. Yaml comment.** Updated to describe env-var substitution rather than an override path.

### Validation

- `go test ./... -race -count=1` — green (incl. new `rpc_redact_test.go` + 4 new `loader_test.go` tests)
- `cargo test --workspace --release` — 395 passed
- `forge test` — 38 passed, 2 skipped
- `go vet ./...` / `go build ./...` — clean
- `cargo clippy --workspace --release -- -D warnings` — clean
- Fail-fast smoke tests:
  - no `ETH_RPC_URL` → `FATAL: ETH_RPC_URL not set …`
  - no `AETHER_EXECUTOR_ADDRESS` → `FATAL: … executor_address must not be empty`
  - zero address → `FATAL: … must not be the zero address`
  - unreachable RPC (`http://127.0.0.1:1`) → fatal with redacted URL
  - bogus Alchemy key → `FATAL: eth_chainId failed: 401 Unauthorized …` (no key in log)

Ready for another look.

internal/config/loader.go
```diff
@@ -224,6 +236,64 @@ type NodesFileConfig struct {
...
+	}
+
+	return cfg, nil
+}
+
+// ValidateExecutorConfig ensures required fields are present and well-formed.
+// Address syntax is validated here; on-chain bytecode check happens at runtime
+// against the connected Ethereum node.
+func ValidateExecutorConfig(cfg ExecutorFileConfig) error {
+	addr := strings.TrimSpace(cfg.ExecutorAddress)
```
Pablosinyores, line 281: Trims but does not normalize. Consider assigning back (`cfg.ExecutorAddress = addr`) so downstream callers and log sites see a clean value.

cmd/executor/main.go
```diff
@@ -182,12 +225,20 @@ func main() {
...
-	if ethClient != nil && txSigner != nil {
+	// Start ETH balance watcher. The initial fetch is synchronous and fatal
+	// on failure: LiveBalance starts at zero, and risk.PreflightCheck rejects
+	// anything below MinETHBalance, so a transient startup blip would
+	// silently kill every arb for up to 30s until balanceWatchLoop's first
+	// tick. This matches the fatal-on-startup pattern of the dial, chain-ID,
+	// and bytecode checks.
+	if txSigner != nil {
+		if err := fetchAndStoreBalance(ctx, ethClient, txSigner.Address(), liveBalance); err != nil {
+			log.Fatalf("FATAL: initial eth_getBalance failed: %v", redactRPCError(err, rpcURL))
```
Pablosinyores: Including the address in this fatal message would make incident triage easier. The earlier `Searcher address:` log line has the address, but a self-contained fatal is nicer when all you have is the crash line.

Pablosinyores (approved):
## Re-review after fixes

All 8 items from the prior review are resolved in `fix(executor): address PR #81 review feedback`. The RPC URL redaction (`rpc_redact.go`) is the right abstraction, test coverage for hex validation and env expansion is thorough, the fatal initial balance fetch eliminates the 30s silent arb-rejection window, and the atomic `LiveBalance` pattern is race-clean.

### Non-blocking notes

- `internal/config/loader.go:277` trims whitespace but does not write the trimmed value back into `cfg`. No functional bug (downstream `HexToAddress` tolerates whitespace) but the logged address could show the padding.
- `decodeStrictYAML` is applied to executor config only. Risk/builders/nodes loaders still silently ignore typos, worth a follow-up.
- A direct unit test for `LiveBalance` would make the atomic float64 contract explicit and catch any future refactor that breaks the invariant.

### Verdict: APPROVE

0xfandom:
Addressed the re-review non-blocking notes in `80d356f`:

**1. `internal/config/loader.go` — trim write-back.**
`LoadExecutorConfig` now assigns the trimmed `ExecutorAddress` back into `cfg` before validation, so the returned config and all downstream log sites see the normalized value. `ValidateExecutorConfig` keeps its internal defensive `TrimSpace`; the new assignment just persists it.

**2. `cmd/executor/main.go:236` — address in fatal.**
The initial `eth_getBalance` fatal log now includes the searcher address:
```
FATAL: initial eth_getBalance(0x...) failed: <redacted-error>
```
Self-contained for incident triage; the redacted error path is unchanged.

**3. `cmd/executor/metrics_test.go` — direct `LiveBalance` unit tests.**
Four new tests pin the atomic-float64 contract so a future refactor that silently swaps `atomic.Uint64` for a plain `float64` (which would tear reads under contention) fails loudly:
- `TestLiveBalance_ZeroValue` — fresh instance reads 0
- `TestLiveBalance_SetGetRoundTrip` — exact round-trip across edge values (`0`, `1e18`, `1e-18`, `MaxFloat64`, `SmallestNonzeroFloat64`)
- `TestLiveBalance_OverwritesPreviousValue` — last write wins
- `TestLiveBalance_ConcurrentSetGet` — 8 writers × 8 readers × 10,000 iters, asserts every observed read matches a value that was written (no torn bit patterns)

**4. Strict YAML decoding on risk / builders / nodes loaders — deferred.**
That one touches three unrelated config contracts and deserves its own test coverage per loader; filing as a follow-up rather than folding into this PR.

### Validation

- `go test ./... -race -count=1` → all packages pass (`cmd/executor`, `cmd/pooldiscovery`, `internal/config`, `internal/risk`)
- `go vet ./...` → clean
- Targeted re-run: `go test -run "TestLiveBalance|TestValidateExecutorConfig|TestLoadExecutorConfig" -race -count=1 ./...` → all green

PR #85, opened 2026-04-14
feat(grpc-server): surface decoder drops via warn log + counter

### LINKED ISSUE #67: [E1/WS-1] Replace event decoder panics with Result types
## Context

`crates/ingestion/src/event_decoder.rs` contains 6 `panic!("Expected X event")` calls at lines 284, 305, 370, 401, 445, 491. These crash the entire Rust process on any malformed or unexpected log event — a single bad event from a misbehaving pool kills the hot path.

`crates/ingestion/src/subscription.rs` has the same pattern at lines 152 and 234.

### WS-1.4 — event_decoder.rs
- Define `DecodeError::UnexpectedEvent { expected: &'static str, got: String }` in `crates/common/src/error.rs`
- Replace all 6 `panic!()` calls with `return Err(DecodeError::UnexpectedEvent { ... })`
- Propagate errors to the subscription loop caller
- In the subscription loop: increment `aether_decode_errors_total` counter, log at `warn` level, and **skip the event** (do not crash)

### WS-1.5 — subscription.rs
- Same treatment for lines 152 and 234
- Replace `panic!`/`unwrap` with proper `Result` propagation
- Subscription loop must continue processing remaining events after a decode error

## Files

- `crates/ingestion/src/event_decoder.rs` — 6 panic sites → Result
- `crates/ingestion/src/subscription.rs` — 2 panic sites → Result
- `crates/common/src/error.rs` — add `DecodeError::UnexpectedEvent` variant
- `cmd/monitor/metrics.go` or Rust metrics — register `aether_decode_errors_total` counter

## Acceptance Criteria

- [ ] `grep -r 'panic!("Expected' crates/ingestion/` returns zero matches
- [ ] `grep -rn 'unwrap()' crates/ingestion/src/event_decoder.rs` returns zero matches on decode paths
- [ ] New `DecodeError::UnexpectedEvent` variant exists with descriptive fields
- [ ] Malformed event logs a warning and increments `aether_decode_errors_total` — does not crash
- [ ] `cargo test --workspace --release` passes
- [ ] `cargo clippy --workspace -- -D warnings` clean

## Epoch

E1 — Stop the Bleeding. This is the highest-severity production risk in the Rust core.

0xfandom:
## Context — issue premise audit

This PR addresses the *real* observability gap behind #67. See [this comment on #67](https://github.com/Pablosinyores/aether/issues/67#issuecomment-4243288445) for the full audit.

Short version:
- All eight cited `panic!` sites (event_decoder.rs 284/305/370/401/445/491, subscription.rs 152/234) lived inside `#[cfg(test)] mod tests` — idiomatic Rust unit-test assertion failures, not production panics.
- Production `event_decoder::decode_log` already returns `Option<PoolEvent>` — malformed events already don't crash the process.
- The real consumer is `crates/grpc-server/src/provider.rs:437,455` (not `subscription.rs`). Both call sites handled the `None` arm but **silently dropped** the event — no log, no metric, no alert.

## Summary (shipped design — post review round 1)

- New `aether_decode_errors_total` counter on `EngineMetrics`.
- `RpcProvider::new(config, channels, metrics: Arc<EngineMetrics>)` — **metrics handle is required at construction**, not attached via an optional builder. The `with_metrics(...)` builder from the initial design was dropped after review flagged it as a silent-dead footgun (one refactor-removal of `.with_metrics(...)` would ship the counter at zero forever). Every test constructor uses a shared `test_metrics()` helper.
- Both `decode_log` call sites switch from `if let Some` to `match`; on `None` we now increment the counter and emit a `trace!` with pool address + topic0 via `record_decode_failure()`.
- `trace!` instead of `warn!` at the per-event site — pool-discovery mode routinely receives thousands of non-decodable events per block (unregistered ERC20 transfers, NFT events, etc.); a `warn!` per event would swamp Loki. The counter is the alert surface; `trace!` is only for targeted debugging.
- Module visibility: `pub(crate) mod metrics` in `lib.rs` — not `pub mod` — so `provider.rs` can reference it without widening the crate's public API surface.
- Happy path unchanged.

## Test coverage

- `test_process_logs_decode_failure_increments_counter` (new, in `provider.rs`) — builds a `RpcProvider` with a real `EngineMetrics` handle, feeds it a malformed log, asserts `aether_decode_errors_total 1` appears in the rendered Prometheus output. End-to-end coverage of the intended path.
- `test_metrics_render_contains_required_names` strengthened to assert the counter value, not just the metric name.
- The 8 test-site `match { Variant => ..., other => panic!("Expected ...") }` blocks are rewritten as `let Variant { .. } = other else { panic!("...: {:?}", other) }`. Debug info is preserved in the panic message on wrong-variant failures. The literal string `panic!("Expected` no longer appears in the crate (satisfies the #67 grep criterion honestly).

## Results

- `cargo test -p aether-ingestion -p aether-grpc-server` — 239 tests pass.
- `cargo clippy -p aether-ingestion -p aether-grpc-server -- -D warnings` — clean.
- `aether_decode_errors_total` added to the Key Observability Metrics table in `CLAUDE.md` with the alert threshold documented.

## Files Changed

| File | Change |
|---|---|
| `crates/grpc-server/src/metrics.rs` | new `decode_errors: IntCounter` field + `inc_decode_errors()`, registered in Prometheus registry; `Default` impl added; test asserts `aether_decode_errors_total` name + value |
| `crates/grpc-server/src/provider.rs` | `metrics: Arc<EngineMetrics>` now required on `RpcProvider::new`; `record_decode_failure()` helper (counter + `trace!`); `test_metrics()` helper for test construction; new end-to-end test for the decoder-drop path |
| `crates/grpc-server/src/lib.rs` | `pub(crate) mod metrics` — crate-internal, not a public item |
| `crates/grpc-server/src/main.rs` | constructs `Arc<EngineMetrics>` once, passes it to `RpcProvider::new` |
| `crates/ingestion/src/event_decoder.rs` | 6 test-site `match`-then-`panic` rewritten as `let-else` with `{:?}` Debug info |
| `crates/ingestion/src/subscription.rs` | 2 test-site `match`-then-`panic` rewritten as `let-else` with `{:?}` Debug info |
| `CLAUDE.md` | `aether_decode_errors_total` added to Key Observability Metrics table with alert threshold |

## Acceptance Criteria

- [x] Silent-drop of malformed events surfaces via a Prometheus counter that alert authors can reference by name.
- [x] Counter is non-optional on the provider — cannot be disabled by a future refactor.
- [x] Pool-discovery log volume doesn't swamp Loki (downgrade to `trace!`).
- [x] Test asserts counter value on the end-to-end decode-failure path.
- [x] `panic!("Expected ...` literal no longer appears outside of `{:?}` Debug-inclusive messages.
- [x] Metric name is documented in `CLAUDE.md` as the single source of truth.

## Test plan

- [x] `cargo test -p aether-ingestion -p aether-grpc-server` — 239/239 pass
- [x] `cargo clippy -p aether-ingestion -p aether-grpc-server -- -D warnings` — clean
- [x] New test `test_process_logs_decode_failure_increments_counter` asserts both counter presence and value

Closes #67.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -305,6 +305,7 @@ aether/
 | `aether_gas_price_gwei` | Gauge | >300 → halt |
 | `aether_daily_pnl_eth` | Gauge | <-0.5 ETH → halt |
 | `aether_eth_balance` | Gauge | <0.1 ETH → halt |
+| `aether_decode_errors_total` | Counter | sustained >10/min → warn (ABI drift or malformed event floods) |
 
 ---
 
--- a/crates/grpc-server/src/engine.rs
+++ b/crates/grpc-server/src/engine.rs
@@ -25,8 +25,9 @@ use aether_state::snapshot::SnapshotManager;
 use aether_state::token_index::TokenIndex;
 
 // Import the proto ValidatedArb type from service module
+use aether_grpc_server::EngineMetrics;
+
 use crate::pipeline;
-use crate::metrics::EngineMetrics;
 use crate::service::aether_proto::ValidatedArb as ProtoValidatedArb;
 
 /// Configuration for the AetherEngine.
--- a/crates/grpc-server/src/lib.rs
+++ b/crates/grpc-server/src/lib.rs
@@ -2,6 +2,10 @@
 ///
 /// Re-exports the `provider` module so that integration tests and other
 /// crates can access `ProviderConfig`, `RpcProvider`, and related types
-/// without depending on the binary entry point.
-pub mod metrics;
+/// without depending on the binary entry point. The `metrics` module is
+/// crate-private; only the two types the binary and integration tests
+/// actually need are re-exported publicly.
+pub(crate) mod metrics;
 pub mod provider;
+
+pub use metrics::{start_metrics_server, EngineMetrics};
--- a/crates/grpc-server/src/lib.rs
+++ b/crates/grpc-server/src/lib.rs
@@ -3,4 +3,5 @@
 /// Re-exports the `provider` module so that integration tests and other
 /// crates can access `ProviderConfig`, `RpcProvider`, and related types
 /// without depending on the binary entry point.
+pub mod metrics;
 pub mod provider;
--- a/crates/grpc-server/src/main.rs
+++ b/crates/grpc-server/src/main.rs
@@ -13,9 +13,10 @@ use tokio::net::UnixListener;
 use tokio_stream::wrappers::UnixListenerStream;
 
 mod engine;
-mod metrics;
 mod pipeline;
 mod service;
+
+use aether_grpc_server::metrics;
 
 use aether_grpc_server::provider::{ProviderConfig, RpcProvider};
 use engine::{AetherEngine, EngineConfig};
@@ -92,10 +93,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     if provider_config.nodes_config_path.is_some() {
         info!("AETHER_NODES_CONFIG set — provider will use multi-node pool");
     }
-    let provider = Arc::new(RpcProvider::new(
-        provider_config,
-        Arc::clone(engine.event_channels()),
-    ));
+    let provider = Arc::new(
+        RpcProvider::new(provider_config, Arc::clone(engine.event_channels()))
+            .with_metrics(Arc::clone(&metrics)),
+    );
 
     // Shutdown coordination via watch channel.
     let (shutdown_tx, shutdown_rx) = tokio::sync::watch::channel(false);
--- a/crates/grpc-server/src/main.rs
+++ b/crates/grpc-server/src/main.rs
@@ -16,11 +16,9 @@ mod engine;
 mod pipeline;
 mod service;
 
-use aether_grpc_server::metrics;
-
 use aether_grpc_server::provider::{ProviderConfig, RpcProvider};
+use aether_grpc_server::{start_metrics_server, EngineMetrics};
 use engine::{AetherEngine, EngineConfig};
-use metrics::{start_metrics_server, EngineMetrics};
 use service::aether_proto::arb_service_server::ArbServiceServer;
 use service::aether_proto::control_service_server::ControlServiceServer;
 use service::aether_proto::health_service_server::HealthServiceServer;
@@ -93,10 +91,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     if provider_config.nodes_config_path.is_some() {
         info!("AETHER_NODES_CONFIG set — provider will use multi-node pool");
     }
-    let provider = Arc::new(
-        RpcProvider::new(provider_config, Arc::clone(engine.event_channels()))
-            .with_metrics(Arc::clone(&metrics)),
-    );
+    let provider = Arc::new(RpcProvider::new(
+        provider_config,
+        Arc::clone(engine.event_channels()),
+        Arc::clone(&metrics),
+    ));
 
     // Shutdown coordination via watch channel.
     let (shutdown_tx, shutdown_rx) = tokio::sync::watch::channel(false);
--- a/crates/grpc-server/src/metrics.rs
+++ b/crates/grpc-server/src/metrics.rs
@@ -145,6 +145,12 @@ impl EngineMetrics {
     }
 }
 
+impl Default for EngineMetrics {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 pub fn start_metrics_server(metrics: Arc<EngineMetrics>) {
     let addr = metrics_addr();
 
--- a/crates/grpc-server/src/metrics.rs
+++ b/crates/grpc-server/src/metrics.rs
@@ -134,7 +134,10 @@ impl EngineMetrics {
         self.decode_errors.inc();
     }
 
-    fn render(&self) -> Vec<u8> {
+    /// Render the registered metrics in Prometheus text exposition format.
+    /// `pub(crate)` so sibling modules (`provider::tests`) can assert on
+    /// rendered counter values without exposing the whole registry.
+    pub(crate) fn render(&self) -> Vec<u8> {
         let metric_families = self.registry.gather();
         let encoder = TextEncoder::new();
         let mut buffer = Vec::new();
@@ -269,5 +272,6 @@ mod tests {
         assert!(output.contains("aether_simulations_run_total 3"));
         assert!(output.contains("aether_arbs_published_total 4"));
         assert!(output.contains("aether_blocks_processed_total 1"));
+        assert!(output.contains("aether_decode_errors_total 1"));
     }
 }
--- a/crates/grpc-server/src/metrics.rs
+++ b/crates/grpc-server/src/metrics.rs
@@ -15,6 +15,7 @@ pub struct EngineMetrics {
     simulations_run: IntCounter,
     arbs_published: IntCounter,
     blocks_processed: IntCounter,
+    decode_errors: IntCounter,
 }
 
 impl EngineMetrics {
@@ -57,6 +58,11 @@ impl EngineMetrics {
             "Total blocks processed",
         )
         .expect("aether_blocks_processed_total counter");
+        let decode_errors = IntCounter::new(
+            "aether_decode_errors_total",
+            "Total logs the event decoder could not parse",
+        )
+        .expect("aether_decode_errors_total counter");
 
         registry
             .register(Box::new(detection_latency_ms.clone()))
@@ -76,6 +82,9 @@ impl EngineMetrics {
         registry
             .register(Box::new(blocks_processed.clone()))
             .expect("register aether_blocks_processed_total");
+        registry
+            .register(Box::new(decode_errors.clone()))
+            .expect("register aether_decode_errors_total");
 
         Self {
             registry,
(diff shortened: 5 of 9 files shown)


Pablosinyores (changes requested):
Pulled the branch and ran through it end-to-end. The audit in the PR body is correct — verified all 8 original `panic!` sites on `origin/main` (`event_decoder.rs:284,305,370,401,445,491` and `subscription.rs:152,234`) sit below the `#[cfg(test)]` markers at line 223 / 103 respectively, and the PR body's claim that production `decode_log` has always returned `Option<PoolEvent>` checks out. No production panic regression. Tests and clippy are clean: `aether-ingestion` 66/66 pass, `aether-grpc-server` 20/20 pass, `cargo clippy -p aether-ingestion -p aether-grpc-server -- -D warnings` green and matching the main baseline.

That said, the headline feature — "surface decoder drops via counter" — has a few gaps that should be closed before merge.

### Needs fixing

**1. `with_metrics` builder is a silent-dead footgun.** `crates/grpc-server/src/provider.rs:69,113-116,479-481` — the `metrics` field is `Option<Arc<EngineMetrics>>` attached via a builder. `main.rs:96-99` does call `.with_metrics(...)` today, but nothing in the type system enforces it. Drop that one line during a future refactor and the feature still compiles, every test passes, and it ships with a dead counter forever. Counter-at-zero doesn't alert the same way missing-series does. Suggest making the field non-`Option` and requiring it at construction: `RpcProvider::new(config, channels, metrics)`. Test code can pass `Arc::new(EngineMetrics::new())`.

**2. No end-to-end test for the decoder-drop path.** The only test that exercises `inc_decode_errors()` is in `metrics.rs` and calls it directly on the struct. `provider.rs::test_process_logs_unknown_event_ignored` builds a provider *without* metrics attached. Even `test_metrics_render_contains_required_names` only asserts the metric *name* appears in the `/metrics` output (line 258) — it doesn't check the value incremented. Contrast the histogram assertions at 264-271, which do check values. Add something like:

```rust
#[test]
fn process_logs_decode_failure_increments_counter() {
    let metrics = Arc::new(EngineMetrics::new());
    let provider = RpcProvider::new(cfg, chans).with_metrics(Arc::clone(&metrics));
    provider.process_logs(&[bad_log]);
    assert!(String::from_utf8(metrics.render()).unwrap()
        .contains("aether_decode_errors_total 1"));
}
```

Strengthen the existing render test while you're there — assert `aether_decode_errors_total 1`, not just that the name appears.

**3. `warn!` per dropped log will flood in pool-discovery mode.** `provider.rs:478-488` emits a `warn!` on every `None` return — no sampling, no rate limit. In discovery mode where `monitored_pools = []`, every unmatched log on mainnet (tens of thousands per block) fires a warn line. Loki bytes-per-second limits will either drop lines or backpressure the process. Downgrade per-event to `trace!`, or sample (every 1000th drop). The counter alone already gives ops the signal; the log is for triage and doesn't need to be on every event.

### Architectural, worth discussing

**4. The counter collapses three genuinely different failure modes.** Reading through `decode_log` in `event_decoder.rs:101-126`:

- Unknown/unhandled topic0 (benign, high volume in discovery mode)
- Malformed payload (`data.len() < 64` in `decode_sync`, `< 160` in `decode_swap_v3`, `< 3 topics` in `decode_pair_created` — this is a real bug that should alert)
- **Silent drops** — `event_topics()` at line 434 includes the V2 `Swap` topic, but `decode_log` has no branch for it. Every V2 Swap on every monitored pool currently drops and is indistinguishable from generic unknown-topic noise.

A single-dimension counter can't separate these. The PR body rejects `DecodeError` as "not needed" but for a production trading system the cost of a structured error (same tagged-union size in practice; `Err(e)` vs `None` is free on the happy path) is negligible and the observability win is real. This doesn't have to block merge, but please file a follow-up to split by reason — and independently, the V2 Swap no-branch case looks like a genuine bug that deserves its own issue.

### Nits

**5.** `let...else` refactor dropped the variant info from panic messages. `event_decoder.rs:272-280` et al. — old form was `panic!("Expected ReserveUpdate, got {:?}", other)`, new form is `panic!("decoder returned unexpected variant")`. Tests still fail correctly on the wrong variant; the regression is only in debuggability. `panic!("decoder returned unexpected variant: {:?}", received)` restores the diagnostic at zero cost.

**6.** `pub mod metrics` in `lib.rs:6` — the publicity is unused outside `provider.rs` today. `pub(crate) mod metrics` would keep the API surface smaller. If you add the integration test from (2) under a `tests/` dir, leave it `pub`.

### Verified good

- Audit of `panic!` sites was accurate (all 8 under `#[cfg(test)]`).
- No production `unwrap()` regressions — every `unwrap()` in `event_decoder.rs` and `subscription.rs` lives under `#[cfg(test)]`.
- `aether_decode_errors_total` is a proper monotonic counter with `_total` suffix; `inc_decode_errors` is a single atomic increment.
- Clippy clean, matching main baseline (commit `7d9476a` addressed the `new_without_default` warning `with_metrics` would have introduced).
- Acceptance criterion #1 (zero `panic!("Expected` matches in `crates/ingestion/`) met.

(2) is the most important one — this PR's reason for existing is adding a metric, and no test today confirms the metric actually increments from the real call path. Happy to re-review once (1) and (2) land; (3) and (4) can go in the same push or split into follow-ups, your call.

0xfandom:
Thanks for the thorough pass. All items addressed in `cea23de`.

### Blockers

**1. `metrics` non-`Option` at construction.** `RpcProvider::new` now takes `Arc<EngineMetrics>` as a required parameter (`crates/grpc-server/src/provider.rs:87`), the field is no longer `Option` (`:71`), and `with_metrics()` is gone entirely. Forgetting to wire metrics is now a compile error, not a silent footgun. 15 call sites updated — 11 in `provider::tests`, 3 in `tests/ws_subscription_test.rs`, 1 in `main.rs`. Tests get `test_metrics()` helpers that return fresh `Arc::new(EngineMetrics::new())` so concurrent tests don't share registries.

**2. End-to-end counter test.** New `test_process_logs_decode_failure_increments_counter` (`provider.rs:649`) drives a malformed log through the real `process_logs → record_decode_failure → inc_decode_errors` path and asserts the rendered Prometheus output contains `aether_decode_errors_total 1`, then asserts `_total 2` after a second drop (catches any accidental reset). Made `EngineMetrics::render()` `pub(crate)` so sibling tests can inspect the output. The existing `test_metrics_render_contains_required_names` is also strengthened (`metrics.rs:275`) to assert the value `aether_decode_errors_total 1`, matching the histogram assertions above it instead of just checking the name.

### Worth fixing

**3. `warn!` → `trace!`.** `record_decode_failure` (`provider.rs:495`) now logs at `trace!`. The counter stays as the primary ops signal; the per-event line is there only for targeted debugging when someone actively raises `RUST_LOG`. Discovery-mode floods no longer reach Loki.

### Architectural

**4.** Both follow-ups filed:
- **#91** — V2 Swap silent drop (the genuine bug): `event_topics()` advertises `swap_v2_topic()` but `decode_log` has no matching branch, so every UniV2 / Sushi Swap silently lands in the `else { None }` arm.
- **#92** — Split `aether_decode_errors_total` by `reason` label so `unknown_topic`, `malformed_payload`, and `insufficient_topics` are triageable separately.

### Nits

**5. Variant diagnostic restored.** All 8 `let...else` sites now bind the unexpected value to `got` / `received` first and interpolate it into the panic message (`event_decoder.rs:280, 303, 365, 398, 438, 486`; `subscription.rs:149, 229`). Tests still fail on the wrong variant; now with the full Debug of the received event for triage.

**6. Visibility narrowed.** `lib.rs` is now `pub(crate) mod metrics` plus targeted `pub use metrics::{start_metrics_server, EngineMetrics}` at the crate root. `main.rs` and `engine.rs` consume the re-exports by type, and `ws_subscription_test.rs` uses `aether_grpc_server::EngineMetrics` directly. The module path itself is no longer part of the public API.

### Validation

- `cargo test -p aether-grpc-server -p aether-ingestion --release` — 173 pass, 0 fail (was 166; +1 E2E test + the new test_metrics helper count).
- `cargo test --workspace --release` — 396 pass, 0 fail.
- `cargo clippy -p aether-grpc-server -p aether-ingestion --release -- -D warnings` — clean, matches main baseline.
- Full `go test -race -count=1 ./...` + `forge test` — clean (not touched by this PR, sanity check only).
- `grep -rn 'panic!("Expected' crates/ingestion/` — still zero matches (Issue #67 acceptance criterion).

Ready for another look.

Pablosinyores (approved):
## Re-review after fixes

All 6 items from the prior review are addressed. Switching from optional builder to required-at-construction closes the silent-dead footgun, the new `test_process_logs_decode_failure_increments_counter` covers the intended path end to end, `warn!` is downgraded to `trace!` with a clear rationale for discovery mode, `let...else` sites restore Debug info in panic messages, and `pub(crate) mod metrics` narrows the surface correctly. Clippy clean, 239 tests pass across grpc-server and ingestion.

### Non-blocking

- The PR description still describes the original design (`with_metrics` builder, `Option<Arc<EngineMetrics>>`, `warn!`, `pub mod metrics`). Please refresh it before squash-merge so the commit message and git log reflect what actually shipped.
- Worth adding `aether_decode_errors_total` to the Key Observability Metrics table in CLAUDE.md so alert authors have a single source of truth.

### Verdict: APPROVE

PR #86, opened 2026-04-14
feat(observability): distributed tracing to tempo + canary probe

Pablosinyores:
## Summary

Closes the observability series for issue #69 (WS-7.6 + WS-7.7). Stacked on PR #84. Adds distributed tracing to a self-hosted Tempo and a scrape-staleness canary so we alert when a service stops emitting metrics altogether, not just when metric values cross thresholds.

## What's in

- **Tempo** service in \`deploy/docker/docker-compose.yml\`, single-binary, filesystem backend, 72h span retention. Listens on OTLP/gRPC 4317 and exposes the query HTTP API on 3200.
- **Grafana Tempo datasource** with \`tracesToLogsV2\` wired to the Loki datasource from PR2 so clicking a span jumps to that service's log lines for the matching time window.
- **Rust OTLP export** (opentelemetry 0.27 + tracing-opentelemetry 0.28). New \`tracing_init\` module builds the Registry in one pass with fmt/json + OTLP layers; disabled when \`OTEL_EXPORTER_OTLP_ENDPOINT\` is unset, so \`cargo run\` stays zero-config.
- **Rust per-stage spans** on the hot path: \`ingest.pool_update\` / \`ingest.new_block\` / \`ingest.pending_tx\` on the dispatcher, \`engine.detection_cycle\` around each cycle with a \`detect\` child span, and \`arb.publish\` with \`arb_id\` when the validated arb is broadcast to the Go executor. Bellman-Ford and revm inner loops stay uninstrumented to keep sampling overhead out of the hot path.
- **Go OTLP export** via \`go.opentelemetry.io/otel/sdk\` + \`otlptracegrpc\`. Helpers at \`cmd/executor/tracing.go\` and \`cmd/monitor/tracing.go\` build an exporter + BatchSpanProcessor when the env var is set, otherwise return a no-op shutdown.
- **Go per-stage spans**: \`processArb\` root span tagged with \`arb_id\` (correlates with the Rust \`arb.publish\` span), plus \`preflight\`, \`bundle.build\`, \`submit.fanout\`, and per-goroutine \`submit.builder\` child spans with per-builder success/failure status.
- **Canary probe** (\`scripts/canary.py\`, stdlib-only). GETs three endpoints every 60s, fires a \`ScrapeStale\` alert to Alertmanager's \`/api/v2/alerts\` API when a target has been down for >180s, and sends a resolved alert on recovery. Packaged as a \`python:3.12-alpine\` container and added to the compose stack as \`canary\`.

## Trace correlation decision

Rust and Go emit **separate root traces** that share the \`arb_id\` attribute. No \`traceparent\` injection through the proto schema; joining across processes happens via Tempo queries on \`arb_id\`. This keeps the PR scoped and avoids a proto change; full distributed-trace joining can land later if the cross-trace UX is insufficient.

## Test plan

- [x] \`go build ./... && go vet ./...\` clean
- [x] \`go test ./cmd/executor/... ./internal/risk/... ./cmd/monitor/... ./cmd/pooldiscovery/...\` green
- [x] \`cargo check --workspace\` clean
- [x] \`cargo test -p aether-grpc-server -p aether-ingestion --lib\` green
- [x] \`docker-compose.yml\`, \`tempo-config.yml\`, \`tempo.yml\` (datasource) all parse as valid YAML
- [x] \`python3 scripts/canary.py --self-test\` prints configured targets and exits 0
- [ ] Bring the stack up, trigger an arb, confirm a Rust \`arb.publish\` span and a Go \`processArb\` span both appear in Grafana Tempo Explore with matching \`arb_id\` attribute.
- [ ] Stop \`aether-go\` for >3min, confirm \`ScrapeStale\` fires in Alertmanager and reaches the Slack receiver from PR1.

## Notes

- Targeting \`feat/observability-pr2\` so the diff stays focused on tracing + canary. Once PR1 and PR2 merge, GitHub rebases this onto main.
- No new secrets. \`OTEL_EXPORTER_OTLP_ENDPOINT\` is wired inside the compose network; \`SLACK_WEBHOOK_URL\` is still the only secret env var.
- Go dependency bump: \`go.opentelemetry.io/otel/sdk\` upgraded from v1.39.0 to v1.43.0 as part of \`go get\` pulling in the exporter.


0xfandom (approved):
## PR3 review (WS-7.6 tracing + WS-7.7 canary)

Reviewed on top of PR2 baseline. Hot-path discipline held, span correlation wired correctly, canary is stdlib-clean and alert-shape correct. A handful of medium/low items worth following up on, none blocking.

### Acceptance criteria

| # | Criterion | Evidence | Status |
|---|---|---|---|
| 1 | Rust hot-path inner loops (Bellman-Ford, revm) stay uninstrumented | `rg 'instrument\|span!\|in_scope' crates/detector crates/simulator` → no matches outside tests | Met |
| 2 | OTLP zero-cost when endpoint unset (Rust) | `crates/grpc-server/src/tracing_init.rs:38-61` builds exporter + provider only when `OTEL_EXPORTER_OTLP_ENDPOINT` is non-empty; otherwise `otel_layer=None` and registry line 85 is effectively a no-op | Met |
| 3 | OTLP zero-cost when endpoint unset (Go) | `cmd/executor/tracing.go:28-31`, `cmd/monitor/tracing.go:20-23` early-return no-op shutdown; `otel.SetTracerProvider` is only called on the success path so `tracer.Start(...)` stays on the global no-op provider | Met |
| 4 | Span flush on shutdown | Rust `TracingGuard::drop` calls `global::shutdown_tracer_provider()` (`tracing_init.rs:15-20`); Go `main.go:112-118` / `cmd/monitor/metrics.go:110-115` defer `tp.Shutdown` under 5s timeout | Met |
| 5 | arb_id correlation key matches Rust↔Go | Rust `engine.rs:1266-1271` `info_span!("arb.publish", arb_id = %input.opp.id, ...)`; Go `main.go:275-281` `tracer.Start(ctx, "processArb", ...attribute.String("arb_id", arb.Id))` — same key, same value | Met |
| 6 | Per-goroutine builder span uses parent ctx (not orphaned) | `cmd/executor/submitter.go:138-150` — `tracer.Start(ctx, "submit.builder", ...)` inside goroutine uses outer `ctx` captured from `SubmitToAll`, which itself was started from the `processArb` ctx at line 320 | Met |
| 7 | Canary stdlib-only | `scripts/canary.py:28-37` imports only `json`, `os`, `signal`, `sys`, `time`, `urllib.*`, `dataclasses`, `datetime`, `typing` | Met |
| 8 | Alertmanager payload shape | `scripts/canary.py:103-120` emits `labels`, `annotations`, `startsAt`, `endsAt` (resolved only) — matches `/api/v2/alerts` contract; alert is posted as a JSON array per spec | Met |
| 9 | Resolved path actually resolves | `canary.py:119-120` sets `endsAt = now` when `resolved=True` (line 168-173 path) — Alertmanager interprets `endsAt<=now` as resolved | Met |
| 10 | Slack-only policy preserved | No PagerDuty/Discord references added; canary just POSTs to Alertmanager which routes via PR1's Slack receiver | Met |
| 11 | Tempo + Grafana datasource wiring | `deploy/docker/tempo/tempo-config.yml:16-18` 72h retention; `grafana/provisioning/datasources/tempo.yml:13` tracesToLogsV2 → `datasourceUid: loki` (matches PR2's pinned UID) | Met |
| 12 | Rust OTel version alignment | `Cargo.toml:45-49` — `opentelemetry 0.27` + `tracing-opentelemetry 0.28` is the correct pairing | Met |
| 13 | Go go.mod bump is clean | `go.mod:42-49` — only OTel stack bumped to v1.43.0; no unrelated network/auth libs | Met |
| 14 | No AI/Claude branding | grep of added files — none present | Met |

### Findings

#### MEDIUM — Tempo OTLP gRPC port is published on the host
`deploy/docker/docker-compose.yml:142-144` publishes both `3200:3200` and `4317:4317`. Port 4317 (OTLP ingest) only needs to be reachable on `aether-net` for aether-rust / aether-go / canary. Publishing it to the host means anything that can reach the Docker host can push arbitrary spans into Tempo with no auth — a cheap denial-of-wallet / log-flooding vector, and in a co-located NY5 deployment not a hypothetical one. Suggest dropping the host-side `4317:4317` mapping (leave `3200` for Grafana debug if desired, or drop both since Grafana talks via service DNS).

#### MEDIUM — Per-event spans on the dispatcher with AlwaysOn sampler
`crates/ingestion/src/subscription.rs:64,73,82` wrap `dispatch_pool_update` / `dispatch_new_block` / `dispatch_pending_tx` with `#[tracing::instrument]`. With `Sampler::AlwaysOn` (tracing_init.rs:55) every pool event and every pending tx produces a span that gets batched to the exporter. On a busy block (10k+ events per CLAUDE.md target), that is 10k+ spans/block = ~800 spans/sec steady-state just from ingestion. tracing-opentelemetry serialises attributes on the calling task before handing to the batch queue, so this is measurable work on the dispatcher. Two follow-ups to consider (non-blocking):
1. Switch `dispatch_pending_tx` to a ratio sampler or drop the span entirely — pending-tx volume is the largest and the least diagnostically useful at per-event granularity.
2. Document the expected span rate and Tempo ingest budget in the runbook so ops can spot when per-event instrumentation is drowning the real arb spans.

#### LOW — `otlptracegrpc.WithDialOption()` called with no arguments
`cmd/executor/tracing.go:45`, `cmd/monitor/tracing.go:37`. The option is variadic over `grpc.DialOption` and called with zero args, making it a no-op. Either remove it or pass the dial options you actually want (e.g., `grpc.WithBlock()` if you want the dial to fail-fast instead of async-reconnect). Harmless as-is.

#### LOW — Dial ctx lifetime vs. deferred cancel
`cmd/executor/tracing.go:39-47` — `dialCtx` is cancelled the instant `initTracer` returns because of `defer cancel()`. This is fine as long as `otlptracegrpc.New` returns only after the connection is usable, which the current exporter does. Flagging so future maintainers don't accidentally hand `dialCtx` to something async.

#### LOW — Canary defaults hardcoded to compose service names
`scripts/canary.py:40-44` — defaults point at `aether-rust:9092/metrics` etc. All three are env-overridable via `CANARY_TARGETS`, and compose sets them explicitly at `docker-compose.yml:158`, so this is fine in practice. Worth a line in the canary docstring that the defaults are compose-network-internal and will not resolve on a bare host.

#### LOW — Canary stale threshold hardcoded at 180s
Hardcoded default in `canary.py:72`, overridable via `STALE_AFTER_SEC`. 180s = 3× the 60s probe interval, which is a reasonable pad (tolerates one missed probe) but is unrelated to the Prometheus scrape interval. If the team wants scrape-parity thresholds later, add a doc note that `STALE_AFTER_SEC` should be ≥ 2× the largest metric scrape interval.

#### NIT — Alertmanager outage behaviour is silent-retry
`canary.py:135-139` prints to stderr and drops the alert. Container keeps running on the next cycle, which is the right call — just worth a runbook line that 'canary can't reach Alertmanager' is itself invisible unless you tail the canary container logs. Consider a Prometheus self-scrape of the canary (exposes last-alert-attempt-success as a gauge) as a follow-up.

#### NIT — `Closes #69` position in PR body
Per team convention the `Closes` line should be the last line of the PR body; it is currently in the opening Summary. Not blocking since GitHub parses it regardless.

#### NIT — tracesToLogsV2 tag mapping
`grafana/provisioning/datasources/tempo.yml:18-20` only maps `service.name` → `service`. Once per-arb correlation lands, consider adding `arb_id` to the tag list so the Tempo → Loki jump from a `processArb` span automatically filters Loki logs on that arb — it's the highest-value operator shortcut this stack can provide.

### What is good

- Hot-path discipline: detector and simulator inner loops remain completely uninstrumented, exactly as the PR body claims. Span boundaries sit at `detection_cycle` / `detect` / `arb.publish` — the right granularity.
- The arb_id-based correlation is the pragmatic call: no proto schema churn, TraceQL can join across the two root traces, and the explicit scope-reduction note in the PR body means nobody will be surprised.
- Go no-op path is implemented correctly — `otel.SetTracerProvider` is only invoked when the exporter is up, so `tracer.Start` on the global default stays a true no-op. Many teams get this wrong by installing a real SDK provider with no exporter and paying the span-allocation cost for nothing.
- `TracingGuard` + `defer tp.Shutdown(5s)` is the right shape — spans flush on clean shutdown, bounded at 5s so container stop doesn't hang.
- Canary's Alertmanager v2 payload is spec-correct (labels, annotations, startsAt, endsAt-only-on-resolve, JSON array). Resolved path was easy to get wrong and is right here.
- Dependency bumps are tight: only the OTel stack moved, no transitive bloat in go.sum.

### Verdict

**APPROVE**. Ship it. The two MEDIUMs (Tempo port exposure, per-event span rate) are worth a follow-up issue but neither blocks the series. Nice close-out on E2.

0xfandom:
## Re-review round 2 — hot-path overhead removed + cleanup

Rebased onto current `main` (transitively through updated PR #83 + PR #84). Fix commit: `62a0588`.

### High — hot-path overhead

| # | Item | Status |
|---|---|---|
| 1 | `#[tracing::instrument(skip_all, name = "ingest.*")]` on `dispatch_pool_update`/`dispatch_new_block`/`dispatch_pending_tx` in `subscription.rs` materializes a span through `tracing_subscriber::Registry` for every decoded event, even with OTLP disabled. At 10k events/block bursty mainnet, ~100-300 ns/span × 10k = **1-3 ms/block of bookkeeping cost** with zero consumer when `OTEL_EXPORTER_OTLP_ENDPOINT` is unset | **Fixed in `62a0588`.** Changed all 3 attributes to `#[tracing::instrument(level = "debug", skip_all, name = "ingest.*")]`. At the default `RUST_LOG=info` filter, `EnvFilter` returns `Interest::never()` and the instrument macro compiles to a near-zero-cost interest check. Operators who want these spans can set `RUST_LOG=debug`. Per-block spans in `engine.rs` are untouched — they fire once per block, not per event, so the cost is trivial. |
| 2 | `let _entered = publish_span.enter();` at `engine.rs` `arb.publish` site is an async footgun — if anyone adds an `.await` inside the block later, the span carries across await points (tokio can reschedule across threads while the span is "current"). `broadcast::Sender::send` is sync, so the bug is latent, not current | **Fixed in `62a0588`.** Replaced with `publish_span.in_scope(\|\| { ... })`. Also added `hops = input.opp.hops.len()` to the span attributes for parity with the Go-side `processArb` span (which already recorded `hops`). |

### Medium

| # | Item | Status |
|---|---|---|
| 3 | `cmd/executor/tracing.go` (70 lines) and `cmd/monitor/tracing.go` (62 lines) were byte-for-byte duplicates except for a doc comment — guaranteed to drift the moment anyone adds a propagator / sampler / retry option to one and forgets the other | **Fixed in `62a0588`.** Extracted to `internal/tracing/tracing.go`. Deleted both duplicates. `cmd/executor/main.go` + `cmd/monitor/metrics.go` now call `tracing.Init(ctx, "aether-executor")` / `tracing.Init(ctx, "aether-monitor")`. |
| 4 | The extracted file was using `otlptracegrpc.WithTLSCredentials(insecure.NewCredentials())` + a no-op `otlptracegrpc.WithDialOption()` — works by accident of the TransportCredentials interface but isn't the documented idiom and is fragile across OTel SDK bumps | **Fixed in `62a0588`.** Replaced with `otlptracegrpc.WithInsecure()`. Dropped the `credentials/insecure` import. |
| 5 | Canary only POSTed the firing alert on the state transition `not firing` → `firing`. If Alertmanager is unreachable during that POST (the exact correlated-outage failure this canary exists to detect), the alert is silently lost | **Fixed in `62a0588`.** After the stale-detection branch, the probe now also re-asserts `resolved=False` every cycle while `target.firing=True`. Alertmanager's default resolve timeout is 5 min — this keeps the alert live across Alertmanager blips without hammering it. |
| 6 | `OTEL_EXPORTER_OTLP_ENDPOINT` in compose was inconsistent — `aether-rust` had `http://tempo:4317`, `aether-go` had `tempo:4317` (no scheme). The Go SDK strips prefixes so it still worked, but the OTEL spec says full URL | **Fixed in `62a0588`.** Both services now use `http://tempo:4317`. |

### Low

| # | Item | Status |
|---|---|---|
| 7 | `tracing_init.rs` had `.expect("failed to build OTLP span exporter")` — panics the engine at startup if the OTLP endpoint is misconfigured. Go side already treats export failure as "log and continue". Rust should match | **Fixed in `62a0588`.** Replaced `.expect` with a `match` that `eprintln!`s the error and returns `None` for the `otel_layer`. Service starts without tracing rather than crashing. |

### Deliberately deferred (noise or separate concerns)

- Tempo `metrics_generator` configured without any `processors` — either enable or remove. Cosmetic; belongs in a config-tidy follow-up.
- Canary sequential probe drift under hung targets (would matter only if interval is dropped from 60s).
- Canary unused imports (`field`, `typing.Dict`/`List` legacy forms).
- Version string hardcoded in Go vs `CARGO_PKG_VERSION` in Rust.
- `semconv` pinning version chase.

### Validation

- `go test ./... -race -count=1` — pass (5 packages including new `internal/tracing`)
- `cargo test --workspace --release` — 357 tests pass across 25 binaries
- `cargo clippy --workspace --release --bins --tests -- -D warnings` — clean (main's `8460272 chore(clippy): fix 12 workspace lint errors` is transitively inherited)
- `python3 scripts/canary.py --self-test` — OK
- All tempo / docker-compose YAMLs parse cleanly

Ready for re-review.

0xfandom (commented):
## Summary

PR #86 lands the WS-7.6 + WS-7.7 observability slice cleanly: Tempo service in compose with 72h retention, Rust + Go OTLP exporters, per-stage spans through ingestion/detection/publish on the Rust side and processArb/preflight/bundle.build/submit.fanout/submit.builder on the Go side, and a pure-stdlib canary probe that push-alerts Alertmanager on scrape staleness. All six round-2 fixes verified in place. Hot-path cost with OTEL off is effectively free (dispatcher spans at `level="debug"` + default `info` filter → `Interest::never()`). Rust OTel version triple (`opentelemetry 0.27` / `tracing-opentelemetry 0.28` / `opentelemetry_sdk 0.27`) is the correct compatibility matrix. The deliberate "separate root traces joined by `arb_id`" design is documented and sound for this scope.

## Issue #69 WS-7.6 + WS-7.7 AC table

| AC | Status | Evidence |
|---|---|---|
| Tempo service w/ OTLP ingest + 72h retention | Met | `docker-compose.yml:138-147` (image `grafana/tempo:2.6.1`, ports 3200/4317); `tempo-config.yml:16-18` (`block_retention: 72h`) |
| Rust OTLP span export | Met | `crates/grpc-server/src/tracing_init.rs:49-71` (builder + Tokio batch exporter + ParentBased AlwaysOn) |
| Go OTLP span export | Met | `internal/tracing/tracing.go:46-72` (`otlptracegrpc.WithInsecure()` + `WithBatcher`) |
| Spans on ingestion → detection → publish → bundle → submit | Met | Rust: `subscription.rs:64,73,82`; `engine.rs:741,774-775,1267-1273`. Go: `main.go:330-336,346,356`; `submitter.go:111-114,138-140,144-149` |
| Bellman-Ford + revm inner loops not instrumented | Met | Only `detect` wrapper span around `BellmanFord` calls; no spans inside `crates/detector/src/bellman_ford.rs` or simulator inner loops |
| Canary script + ScrapeStale alert | Met | `scripts/canary.py:180-201` (fire on age≥threshold, re-assert every cycle, resolve on recovery); `canary.Dockerfile:1-7` (minimal image, `USER nobody`) |
| Grafana Tempo datasource wired to Loki + Prometheus | Met | `grafana/provisioning/datasources/tempo.yml:12-30` (`tracesToLogsV2.datasourceUid: loki`, `serviceMap.datasourceUid: prometheus`) |

## Round-2 verification

1. **Dispatcher spans at DEBUG** — Verified. `crates/ingestion/src/subscription.rs:64,73,82` all three dispatchers carry `level = "debug"`. At default `EnvFilter::new("info")` the macro's callsite-interest returns `Interest::never()` and no span is materialized.
2. **`publish_span.in_scope()`** — Verified. `engine.rs:1267-1287` uses `publish_span.in_scope(|| { ... })`, no `.enter()`. Fields `arb_id`, `hops = input.opp.hops.len()`, `net_profit_wei`, `sim_us` all present (line 1269-1272).
3. **`internal/tracing/` extraction** — Verified. File exists at `internal/tracing/tracing.go:1-73`. `cmd/executor/tracing.go` and `cmd/monitor/tracing.go` both deleted. `cmd/executor/main.go:24` and `cmd/monitor/metrics.go:12` import `"github.com/aether-arb/aether/internal/tracing"`. Init uses `otlptracegrpc.WithInsecure()` (`tracing.go:49`). No `WithTLSCredentials`, no `credentials/insecure` import, no no-op `WithDialOption()` anywhere.
4. **Canary re-assert** — Verified. `scripts/canary.py:181-190` is the original transition path; `:197-201` re-asserts `resolved=False` every cycle while `target.firing` is true.
5. **Fail-open tracing init** — Verified. `crates/grpc-server/src/tracing_init.rs:49-61` uses `match … { Ok(e) => e, Err(err) => { eprintln!(…); return None; } }` — no `.expect`, no panic.
6. **OTEL endpoint scheme consistency** — Verified. `docker-compose.yml:23` (`aether-rust`) and `:44` (`aether-go`) both set `OTEL_EXPORTER_OTLP_ENDPOINT=http://tempo:4317`.

## Must-fix blockers

None.

## Should-fix nits

- **MED — `cmd/executor/main.go:103-104,110` load-bearing reassignment.** `tracer = otel.Tracer("aether-executor")` on line 110 duplicates the package-level `var tracer trace.Tracer = otel.Tracer("aether-executor")` at line 25 — the package-level init happens before any tracer provider is set (`SetTracerProvider` is called inside `Init`), so the line-110 reassignment is what actually captures the real tracer. Worth a comment explaining this is load-bearing, otherwise someone will delete it thinking it's a dupe and silently break tracing.
- **LOW — `tempo-config.yml:28-33` dead `metrics_generator` block.** Block declared but no `overrides.defaults.metrics_generator.processors` configured, so Tempo initializes the module but never routes spans to it. Either delete the block or add `overrides: defaults: metrics_generator: processors: [service-graphs, span-metrics]` if the Grafana `serviceMap` wiring was meant to be functional.
- **LOW — `scripts/canary.py:117-120` `startsAt` rewrite.** Alertmanager preserves first-seen `startsAt` for an already-active alert, so this is harmless semantically, but canary self-logs are misleading on debug.
- **LOW — `canary.py:118,120` `startsAt == endsAt` on resolve.** Some older Alertmanager versions have quirks with `startsAt == endsAt`. Setting `endsAt` to `now - 1s` or leaving `startsAt` at original-fire-time is safer.
- **LOW — `engine.rs:1267-1273` `arb.publish` omits `block_number`.** Go-side `processArb` span has `target_block` (`main.go:334`). Adding `block_number = input.opp.block_number` gives a clean join key across Rust + Go traces.
- **NIT — `tempo.yml` missing `version:` field.** Grafana provisioning tolerates it, but pinning `version: 1` matches the rest of the stack.

## Can-defer

- Cross-process `traceparent` injection through proto (explicit scope-limiting call in PR body — correct decision for this slice).
- Sampling strategy tuning. `ParentBased(AlwaysOn)` is fine at current volume; production high-volume will want tail-based sampling via an OTel Collector in front of Tempo.
- Tempo storage is filesystem-local — acceptable for dev/staging; production will need S3/GCS backend and a separate PR.

## What's good

- Round-2 feedback fully actioned.
- Pure-stdlib canary keeps Python image at ~50MB with no supply-chain surface.
- Rust OTel version triple is the correct compatibility matrix (this ecosystem is a minefield; author picked the right pins).
- `BatchSpanProcessor` on both sides — no synchronous export on the hot path.
- Bellman-Ford and revm inner loops correctly left uninstrumented.

## Verdict

**APPROVE** — ship it. All round-2 fixes verified, ACs complete, no hot-path regressions. The listed nits are post-merge cleanups, not gates.

0xfandom:
## Retargeted to main

Rebuilt on current main after PR #83 and PR #84 (via #103 recovery) landed. The 6 PR #86 commits were cherry-picked onto fresh main; the only conflict was in `crates/grpc-server/src/main.rs` — resolved by taking the `tracing_init::init()` refactor (PR #86's approach) which subsumes the inline JSON layer that PR #84 had set up.

One extra commit added to preserve a PR #84 improvement that would otherwise be lost:

- `6dea247 fix(tracing): preserve .flatten_event(true) after tracing_init refactor` — PR #84 added `.flatten_event(true)` to the inline JSON tracing in `main.rs`. PR #86's `tracing_init.rs` refactor subsumed that block but didn't carry `.flatten_event(true)` forward. Restored so Loki `| json` still exposes fields at the root instead of nested under `fields.*`.

### Validation (post-retarget)

- `go test ./... -race -count=1` — pass
- PR #86 review-round-2 fixes (dispatcher spans at debug, `.in_scope()`, `internal/tracing/` extraction, `WithInsecure()`, canary re-assert, fail-open init, `http://tempo` scheme) — all still in place.

Ready for rebase-and-merge.

Issue #72, opened 2026-04-13
[E4/WS-3] Convert AetherExecutor to UUPS proxy + security fixes + engine-side updates


0xfandom:
## Context

`AetherExecutor.sol` is currently a monolithic immutable contract (7,279 bytes, 30% of 24KB limit). Adding any new DEX requires a full redeployment and config/engine address updates across all services. Phase 2 converts it to a UUPS proxy pattern so the proxy address is permanent and new DEXes are added via implementation upgrades.

Flash loans remain **Aave V3 only** for Phase 2. Balancer (0% fee) flash loans are a future optimization.

Current contract: `contracts/src/AetherExecutor.sol` (332 lines, `solc 0.8.20`).

> **Architecture is open for discussion.** The UUPS proxy approach is the current recommendation based on analysis of all proxy patterns (Transparent, Diamond, Beacon, handler registry). If a better pattern emerges during design review, we should adopt it. The key constraint is zero gas overhead on the hot path.

### WS-3.1 — UUPS Proxy Conversion
- Inherit `Initializable`, `ReentrancyGuardUpgradeable`, `UUPSUpgradeable` from OZ upgradeable
- Split constructor into:
  - `constructor(_aavePool, _balancerVault, _bancorNetwork)` — sets immutables + `_disableInitializers()`
  - `initialize(address _owner)` — sets proxy storage (`owner`, `__ReentrancyGuard_init`, `__UUPSUpgradeable_init`)
- Add `_authorizeUpgrade(address) internal override onlyOwner`
- All existing swap functions (`_swapUniV2/V3/Curve/Balancer/Bancor`) stay as internal functions — unchanged
- Flash loan flow unchanged — Aave V3 `flashLoanSimple` only
- `uniswapV3SwapCallback` stays on the contract — V3 pools call `msg.sender` which is the proxy, proxy delegatecalls to impl, callback validates `_pendingV3Pool` from proxy storage

### WS-3.2 — New Features

**Per-protocol kill switch:**
- `mapping(uint8 => bool) public protocolDisabled`
- `disableProtocol(uint8)` / `enableProtocol(uint8)` — onlyOwner
- Check at top of `_executeSwap()`

**Pause (circuit breaker):**
- `bool public paused` + `whenNotPaused` on `executeArb()`
- `setPaused(bool)` — onlyOwner, callable by Go risk manager

**Two-step ownership:**
- `transferOwnership` → sets `pendingOwner`
[...]

PR #89, opened 2026-04-16
feat(executor): DEX registry + Ownable2Step + security fixes (E4/WS-3 PR1)

Pablosinyores:
Resolves the contract half of #72. Engine-side calldata + `EXECUTOR_ADDRESS` env wiring follow in PR2 (#90, stacked on this branch).

## Why this shape, not the UUPS proxy from the issue

Numeric review showed UUPS costs ~2,940 gas/arb (~$0.31 at 30 gwei) — at the project's 10k arbs/day target that's **~$1.13M/yr** of permanent hot-path tax for the benefit of address permanence and atomic upgrades. Monolith + runtime registry gets most of the flexibility (swap router addresses, kill-switch any protocol) at zero runtime tax; the only thing it gives up is adding a genuinely-new AMM *type* without redeploying, which is rare (≤4×/year expected). Full architecture proof in the approved plan.

## Summary

- **DEX registry**: `protocolRouter` + `protocolEnabled` mappings replace the Balancer/Bancor immutables. `aavePool` stays immutable (flash-loan anchor). Owner can swap routers and kill switches without redeploying.
- **Pre-flashloan validation** (CRITICAL): `executeArb` rejects disabled/unknown protocols before `flashLoanSimple` fires — identified by arch review; avoids burning 150-400k gas on a mid-flashloan revert.
- **Ownable2Step**: two-step ownership replaces manual `owner` + `NotOwner()`. Standard (non-upgradeable) OZ. Incident-response runbook updated for the two-tx rotation.
- **Pause**: Go risk manager can trip a circuit breaker without touching ownership.
- **Security fixes** from WS-3.4: coinbase tip WETH re-wrap fallback, ETH rescue via `token == address(0)`, gasleft() snapshot moved into `executeOperation`, UniV2/Sushi transfer cap via `Math.min(amountIn, balanceOf)`.

## Review-round-2 changes

- **HIGH — drop 2300-gas stipend on `block.coinbase.call`.** Contract-coinbases (splitter/forwarder wrappers) consume more than 2300 in normal operation, silently tripping the WETH fallback and losing builder inclusion priority. Reentrancy is already covered by outer `nonReentrant`; the cap was unnecessary. WETH fallback retained for the "coinbase reverts ETH entirely" case.
- **MEDIUM — push-vs-pull asymmetry on live-balance cap clarified.** Round-1 asked for the UniV2/Sushi cap to extend to Curve/Balancer/Bancor. Initial attempt (`e313b77`) capped `forceApprove` on those three, but round-2 review caught that this only moves the revert from `ERC20InsufficientBalance` to `ERC20InsufficientAllowance` — the pool still parses the amount from `step.data` (encoded off-chain). **Final state** (`88773ec`): `forceApprove(step.amountIn)` restored on the three pull-based protocols, NatSpec rewritten to make the push-vs-pull distinction explicit, engine-side live-balance intersection tracked as part of #97 (the same PR that wires the builders). UniV2/Sushi push-side cap is unchanged and correct.
- **MEDIUM — e2e `setDexRouter(BALANCER_V2, newVault)` routing test added.** The central claim of the registry architecture was previously not exercised end-to-end. New `test_setDexRouter_balancerV2_routesToNewVault` deploys two distinct `CountingBalancerVault` mocks, migrates via `setDexRouter`, and asserts the second vault receives the swap and the first does not.
- Refactor: `_isValidProtocol(uint8) internal pure` helper replaces the duplicated range guards in `setDexRouter` / `setDexEnabled` / pre-flashloan validation. Constructor loop keeps explicit bounds.
- Sentinel test hardcodes numeric IDs (1..=6) so it cannot silently pass if the Solidity constants drift.
- `docs/runbook.md` documents `setPaused` with a `cast send` example, noting the distinction vs. Rust `ControlService/SetState`.
- `docs/incident-response.md` alert-channel references changed PagerDuty/Telegram/Discord → Slack (project convention).

## Numbers

- Deployed bytecode: `AetherExecutor` runtime **8,406 B** — 16,170 B under EIP-170.
- Tests: **59 passed / 0 failed** at `--fuzz-runs 10000` (2 fork tests skipped — need live RPC).
- No runtime gas regression on the hot path (registry SLOAD only on Balancer/Bancor steps; worst-case cold ~2,100 gas absorbed by existing `EXECUTOR_OVERHEAD_GAS = 30k`).

## Test plan

- [x] `forge test --fuzz-runs 10000` — 59/0/2 skipped
- [x] `forge build --sizes` — under EIP-170 limit
- [x] `forge test --match-test test_coinbaseTip -vv` — WETH fallback still trips on reverting coinbase
- [x] `forge test --match-test test_swapUniV2_capsAtBalance` — UniV2 push-side cap still passes
- [ ] Anvil fork end-to-end: deploy → `setDexEnabled(CURVE,false)` → assert pre-flashloan revert → re-enable → arb succeeds → `setDexRouter(BALANCER_V2, newVault)` → arb uses new vault (deferred to integration testing — covered by unit-level routing test)

## Out of scope (follow-ups)

- #90: Rust calldata builders for Curve/Balancer/Bancor + `EXECUTOR_ADDRESS` env var loading.
- #97: Wire the three builders into `engine.rs`, add `pool_id: Option<[u8; 32]>` to `SwapStep`, and intersect `amount_in` with live balance engine-side before encoding — this closes the pull-protocol sizing loop authoritatively.

### DIFF
--- a/contracts/script/Deploy.s.sol
+++ b/contracts/script/Deploy.s.sol
@@ -20,8 +20,8 @@ contract DeployAetherExecutor is Script {
         console.log("AetherExecutor deployed at:", address(executor));
         console.log("Owner:", executor.owner());
         console.log("Aave Pool:", executor.aavePool());
-        console.log("Balancer Vault:", executor.balancerVault());
-        console.log("Bancor Network:", executor.bancorNetwork());
+        console.log("Balancer Vault:", executor.protocolRouter(5)); // BALANCER_V2
+        console.log("Bancor Network:", executor.protocolRouter(6)); // BANCOR_V3
 
         return executor;
     }
--- a/contracts/src/AetherExecutor.sol
+++ b/contracts/src/AetherExecutor.sol
@@ -116,7 +116,7 @@ contract AetherExecutor is Ownable2Step, ReentrancyGuard {
     /// @dev Only valid for BALANCER_V2 and BANCOR_V3 in the current implementation; the
     ///      per-swap-pool protocols keep address(0) here.
     function setDexRouter(uint8 protocol, address router) external onlyOwner {
-        if (protocol == 0 || protocol > BANCOR_V3) revert UnknownProtocol(protocol);
+        if (!_isValidProtocol(protocol)) revert UnknownProtocol(protocol);
         if (router == address(0)) revert ZeroRouter();
         protocolRouter[protocol] = router;
         emit DexRouterSet(protocol, router);
@@ -124,7 +124,7 @@ contract AetherExecutor is Ownable2Step, ReentrancyGuard {
 
     /// @notice Per-protocol kill switch. Idempotent — no event on no-op writes.
     function setDexEnabled(uint8 protocol, bool enabled) external onlyOwner {
-        if (protocol == 0 || protocol > BANCOR_V3) revert UnknownProtocol(protocol);
+        if (!_isValidProtocol(protocol)) revert UnknownProtocol(protocol);
         if (protocolEnabled[protocol] == enabled) return;
         protocolEnabled[protocol] = enabled;
         emit DexEnabledSet(protocol, enabled);
@@ -163,7 +163,7 @@ contract AetherExecutor is Ownable2Step, ReentrancyGuard {
         uint256 stepsLen = steps.length;
         for (uint256 i = 0; i < stepsLen;) {
             uint8 p = steps[i].protocol;
-            if (p == 0 || p > BANCOR_V3) revert UnknownProtocol(p);
+            if (!_isValidProtocol(p)) revert UnknownProtocol(p);
             if (!protocolEnabled[p]) revert ProtocolDisabled(p);
             unchecked { ++i; }
         }
@@ -394,6 +394,12 @@ contract AetherExecutor is Ownable2Step, ReentrancyGuard {
         (bool success,) = network.call(step.data);
         if (!success) revert SwapFailed(index);
         IERC20(step.tokenIn).forceApprove(network, 0);
+    }
+
+    /// @dev Returns true iff `protocol` falls in the valid range [UNISWAP_V2, BANCOR_V3].
+    ///      Zero is reserved as "unset" and is always invalid.
+    function _isValidProtocol(uint8 protocol) internal pure returns (bool) {
+        return protocol >= UNISWAP_V2 && protocol <= BANCOR_V3;
     }
 
     /// @notice Emergency rescue - owner only. token==address(0) rescues native ETH.
--- a/contracts/src/AetherExecutor.sol
+++ b/contracts/src/AetherExecutor.sol
@@ -360,37 +360,59 @@ contract AetherExecutor is Ownable2Step, ReentrancyGuard {
         IERC20(_pendingV3TokenIn).safeTransfer(msg.sender, amountOwed);
     }
 
-    /// @dev Curve: approve pool to pull tokens (capped at live balance), call exchange, reset approval.
-    ///      Capping mirrors the UniV2/Sushi pattern: if the off-chain optimizer over-specs amountIn
-    ///      relative to the live balance the pool's transferFrom would revert mid-flashloan, burning
-    ///      150-400 k gas plus the Aave premium.  The cap makes the on-chain math self-healing.
+    /// @dev Curve: approve pool to pull tokens, call exchange, reset approval.
+    ///
+    ///      PULL-BASED PROTOCOL — the Curve pool calls `transferFrom(executor, pool, amount)`
+    ///      where `amount` is parsed from the calldata encoded in `step.data` (set off-chain
+    ///      by the engine with `step.amountIn`). A live-balance cap at this layer would only
+    ///      reduce the ERC-20 allowance, not the amount encoded in `step.data`. The pool would
+    ///      then attempt `transferFrom` for `step.data`-amount but find allowance < that amount,
+    ///      reverting with ERC20InsufficientAllowance — same gas burn, same Aave premium lost.
+    ///
+    ///      Correct sizing is the engine's responsibility: the off-chain calldata encoder MUST
+    ///      intersect `amount_in` with the live post-prior-hop balance before encoding the swap,
+    ///      ensuring the approved amount and the amount encoded in `step.data` always match.
+    ///      Engine-side live-balance intersection is tracked in issue #97.
+    ///
+    ///      Contrast with UniV2/Sushi (push-based): the executor calls `safeTransfer(pool, actualIn)`
+    ///      before invoking `swap(...)`, so capping the transfer at live balance is safe and correct.
+    ///      That cap remains in place and is unchanged.
     function _swapCurve(SwapStep memory step, uint256 index) internal {
-        uint256 actualIn = Math.min(step.amountIn, IERC20(step.tokenIn).balanceOf(address(this)));
-        IERC20(step.tokenIn).forceApprove(step.pool, actualIn);
+        IERC20(step.tokenIn).forceApprove(step.pool, step.amountIn);
         (bool success,) = step.pool.call(step.data);
         if (!success) revert SwapFailed(index);
         IERC20(step.tokenIn).forceApprove(step.pool, 0);
     }
 
-    /// @dev Balancer V2: approve the registry-configured Vault to pull tokens (capped at live balance).
-    ///      See _swapCurve for the reasoning behind the live-balance cap.
+    /// @dev Balancer V2: approve the registry-configured Vault to pull tokens.
+    ///
+    ///      PULL-BASED PROTOCOL — the Balancer Vault calls `transferFrom(executor, vault, amount)`
+    ///      where `amount` comes from the `IAsset[]` amounts encoded inside `step.data`. A
+    ///      live-balance cap on the allowance alone does not cap the amount the Vault will try
+    ///      to pull — it causes ERC20InsufficientAllowance on any dust shortfall.
+    ///      Amount sizing is the engine's responsibility (see issue #97).
+    ///      See _swapCurve for the full push-vs-pull rationale.
     function _swapBalancer(SwapStep memory step, uint256 index) internal {
         address vault = protocolRouter[BALANCER_V2];
         if (vault == address(0)) revert ZeroRouter();
-        uint256 actualIn = Math.min(step.amountIn, IERC20(step.tokenIn).balanceOf(address(this)));
-        IERC20(step.tokenIn).forceApprove(vault, actualIn);
+        IERC20(step.tokenIn).forceApprove(vault, step.amountIn);
         (bool success,) = vault.call(step.data);
         if (!success) revert SwapFailed(index);
         IERC20(step.tokenIn).forceApprove(vault, 0);
     }
 
-    /// @dev Bancor V3: approve the registry-configured BancorNetwork to pull tokens (capped at live balance).
-    ///      See _swapCurve for the reasoning behind the live-balance cap.
+    /// @dev Bancor V3: approve the registry-configured BancorNetwork to pull tokens.
+    ///
+    ///      PULL-BASED PROTOCOL — BancorNetwork calls `transferFrom(executor, network, sourceAmount)`
+    ///      where `sourceAmount` is ABI-encoded inside `step.data`. A live-balance cap on the
+    ///      allowance alone does not reduce `sourceAmount` in the calldata — it causes
+    ///      ERC20InsufficientAllowance on any shortfall.
+    ///      Amount sizing is the engine's responsibility (see issue #97).
+    ///      See _swapCurve for the full push-vs-pull rationale.
     function _swapBancor(SwapStep memory step, uint256 index) internal {
         address network = protocolRouter[BANCOR_V3];
         if (network == address(0)) revert ZeroRouter();
-        uint256 actualIn = Math.min(step.amountIn, IERC20(step.tokenIn).balanceOf(address(this)));
-        IERC20(step.tokenIn).forceApprove(network, actualIn);
+        IERC20(step.tokenIn).forceApprove(network, step.amountIn);
         (bool success,) = network.call(step.data);
         if (!success) revert SwapFailed(index);
         IERC20(step.tokenIn).forceApprove(network, 0);
--- a/contracts/src/AetherExecutor.sol
+++ b/contracts/src/AetherExecutor.sol
@@ -360,29 +360,37 @@ contract AetherExecutor is Ownable2Step, ReentrancyGuard {
         IERC20(_pendingV3TokenIn).safeTransfer(msg.sender, amountOwed);
     }
 
-    /// @dev Curve: approve pool to pull tokens, call exchange, reset approval
+    /// @dev Curve: approve pool to pull tokens (capped at live balance), call exchange, reset approval.
+    ///      Capping mirrors the UniV2/Sushi pattern: if the off-chain optimizer over-specs amountIn
+    ///      relative to the live balance the pool's transferFrom would revert mid-flashloan, burning
+    ///      150-400 k gas plus the Aave premium.  The cap makes the on-chain math self-healing.
     function _swapCurve(SwapStep memory step, uint256 index) internal {
-        IERC20(step.tokenIn).forceApprove(step.pool, step.amountIn);
+        uint256 actualIn = Math.min(step.amountIn, IERC20(step.tokenIn).balanceOf(address(this)));
+        IERC20(step.tokenIn).forceApprove(step.pool, actualIn);
         (bool success,) = step.pool.call(step.data);
         if (!success) revert SwapFailed(index);
         IERC20(step.tokenIn).forceApprove(step.pool, 0);
     }
 
-    /// @dev Balancer V2: approve the registry-configured Vault to pull tokens.
+    /// @dev Balancer V2: approve the registry-configured Vault to pull tokens (capped at live balance).
+    ///      See _swapCurve for the reasoning behind the live-balance cap.
     function _swapBalancer(SwapStep memory step, uint256 index) internal {
         address vault = protocolRouter[BALANCER_V2];
         if (vault == address(0)) revert ZeroRouter();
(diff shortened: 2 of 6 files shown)


0xfandom (changes requested):
# PR #89 Review — feat(executor): DEX registry + Ownable2Step + security fixes (E4/WS-3 PR1)

Reviewed as PR1 of a 2-PR split. WS-3.5 (Rust calldata builders for Curve/Balancer/Bancor, `ProtocolType` parity, `EXECUTOR_ADDRESS` wiring) and full Anvil integration tests are scoped to PR2 and not flagged here.

## Summary

WS-3.2 (kill-switch/pause/Ownable2Step) and WS-3.4 (4 security fixes) are implemented correctly and well-tested. The architectural pivot from UUPS to a runtime registry is well-justified by the PR body's gas analysis (~$1.13M/yr delegatecall tax) and is accepted as the working scope. Two correctness/security items need attention before merge, plus one governance follow-up that falls out of the UUPS-drop decision.

Tests are strong: 58 tests at `--fuzz-runs 10000`, and the pre-flashloan validation / WETH-fallback suites use real fixtures (`CountingAavePool`, `RevertingCoinbase`) rather than smoke tests.

---

## Findings

### Architecture / Design

**`contracts/src/AetherExecutor.sol:22`, `:40-50`, `:100-117` — Registry replaces two immutables cleanly.**
`balancerVault` / `bancorNetwork` immutables are gone; `protocolRouter[BALANCER_V2]` / `[BANCOR_V3]` are seeded at construction and consumed in `_swapBalancer` / `_swapBancor`. `aavePool` correctly stays immutable — changing the flashloan provider is a protocol-level rearchitecture (different callback selector, different debt-settlement pattern), not a router swap.

**`AetherExecutor.sol:109-115` — constructor seeding.**
```solidity
protocolRouter[BALANCER_V2] = _balancerVault;
protocolRouter[BANCOR_V3]   = _bancorNetwork;
for (uint8 p = UNISWAP_V2; p <= BANCOR_V3; p++) {
    protocolEnabled[p] = true;
}
```
Per-pool protocols (UniV2/V3/Sushi/Curve) deliberately keep `protocolRouter[p] == address(0)` because they don't need a singleton router. NatSpec calls this out.

**🟡 MEDIUM — `setDexRouter` / `setDexEnabled` lack a timelock (`AetherExecutor.sol:121-135`) — consequence of choosing registry over UUPS.**
The PR body lists this as a follow-up. Agreed — but flagging explicitly: these two functions now carry **the same blast radius `_authorizeUpgrade` would have had** under the rejected UUPS design. A compromised owner key can point Balancer trades at a malicious Vault that receives full `forceApprove(amountIn)` every hop. The registry is this architecture's substitute for `upgradeTo`, so it needs equivalent governance (Timelock / short-delay multisig) before mainnet carries material TVL. Not a blocker for this PR; must not ship to production without it.

### Security

**🟠 HIGH — `AetherExecutor.sol:244` — 2300-gas stipend on the coinbase ETH transfer is wrong for contract-coinbases.**
```solidity
(bool sent,) = block.coinbase.call{value: tipAmount, gas: 2300}("");
```
Rationale for the 2300 cap is historically reentrancy, which is a non-issue here (outer `nonReentrant` + Aave settles via post-return `transferFrom`). Real builder contract-coinbases (splitters, forwarders, event-emitting wrappers) consume well over 2300 gas in normal operation, which will **silently trigger the WETH fallback** and:
- costs ~90k extra gas (`withdraw` + `deposit` + `safeTransfer`), and
- delivers WETH — which builders **won't count toward the coinbase-tip inclusion criterion** (the PR's own comment acknowledges "builders won't prioritize").

The fallback test (`test_coinbaseTip_fallsBackToWeth_onRevertingCoinbase`) uses a coinbase that **always reverts**, so it doesn't catch this class of coinbase. Remove the cap:
```solidity
(bool sent,) = block.coinbase.call{value: tipAmount}("");
```
...keeping the existing fallback for the revert case.

**🟡 MEDIUM — `AetherExecutor.sol:278-281, 284-287, 310-333` — the transfer-cap fix is applied only to UniV2/Sushi; Curve/Balancer/Bancor still approve `step.amountIn` unclipped.**
```solidity
if (step.protocol == UNISWAP_V2) {
    uint256 actualIn = Math.min(step.amountIn, IERC20(step.tokenIn).balanceOf(address(this)));
    IERC20(step.tokenIn).safeTransfer(step.pool, actualIn);
    ...
} else if (step.protocol == CURVE) {
    _swapCurve(step, index);   // forceApprove(step.amountIn) — no cap
} else if (step.protocol == BALANCER_V2) {
    _swapBalancer(step, index);
} else if (step.protocol == BANCOR_V3) {
    _swapBancor(step, index);
}
```
`_swapCurve`:
```solidity
IERC20(step.tokenIn).forceApprove(step.pool, step.amountIn);
```
If the off-chain optimizer over-specs `amountIn` relative to live balance on a Curve/Balancer/Bancor hop, `forceApprove` grants the allowance, then the pool's internal `transferFrom` reverts on insufficient balance — killing the whole bundle plus Aave's premium. **Same failure mode the UniV2 cap was added to prevent, just deferred one step.**

Short-term fix: cap the `forceApprove`:
```solidity
uint256 actualIn = Math.min(step.amountIn, IERC20(step.tokenIn).balanceOf(address(this)));
IERC20(step.tokenIn).forceApprove(step.pool, actualIn);
```
Proper fix: PR2's calldata encoders (Curve `exchange(i,j,dx,min_dy)`, Balancer `SingleSwap.amount`, Bancor `tradeBySourceAmount`) also need to respect the same cap, or the encoded `amount` in calldata will drift from the on-chain approval. Flag this when PR2 lands.

**`AetherExecutor.sol:142-154` — pre-flashloan validation is complete.**
```solidity
for (uint256 i = 0; i < stepsLen;) {
    uint8 p = steps[i].protocol;
    if (p == 0 || p > BANCOR_V3) revert UnknownProtocol(p);
    if (!protocolEnabled[p])     revert ProtocolDisabled(p);
    unchecked { ++i; }
}
```
This is the **only** entry path into `executeOperation` (gated on `msg.sender == aavePool` + `initiator == address(this)` at `:159-163`), so no reachable bypass. The redundant `protocolEnabled` check in `_executeSwap:263` is defense-in-depth, correctly rationalised in NatSpec. Both tested.

**`AetherExecutor.sol:290-307` — V3 callback double-spend guard.**
`amountOwed > _pendingV3AmountIn` cap + zero-out-then-transfer is airtight; `test_uniV3Callback_doubleCall_onlyFirstTransfers` exercises it.

**`AetherExecutor.sol:338-345` — ETH rescue.**
Owner-gated; terminal admin call with no post-transfer state that reentry could corrupt. No `nonReentrant` is fine.

**`AetherExecutor.sol:156-158` — gas snapshot relocation.**
`gasStart = gasleft()` now snaps at the top of `executeOperation`, before `msg.sender` / `initiator` checks (~200 gas included). Fine, but the NatSpec claim that this "reflects on-chain work, not calldata encode cost" is slightly off — `abi.decode(params,...)` at `:166` is still counted. Not a correctness bug; comment nit.

**`AetherExecutor.sol:93` — modifier order.**
`onlyOwner nonReentrant whenNotPaused` — owner gate cheapest first, then reentrancy, then paused. Correct. `setPaused` is `onlyOwner`, so the Go risk manager still signs with the owner key to toggle pause — the PR body's "without touching ownership" is technically true (different selector) but mildly misleading on auth.

### Correctness

**`AetherExecutor.sol:166-172` — encode/decode parity.**
Encode: `abi.encode(steps, tipBps, minProfitOut)` (3-tuple). Decode: `abi.decode(params, (SwapStep[], uint256, uint256))`. Consistent — `gasStart` was correctly removed from both sides when it moved into `executeOperation`.

**`AetherExecutor.sol:131-135` — `setDexEnabled` idempotency.**
Returns before writing if state already matches; `test_setDexEnabled_idempotent_noEvent` verifies via `vm.recordLogs()`. Saves 1 SSTORE + avoids spurious events.

**💡 NIT — range-check duplication.**
`protocol == 0 || protocol > BANCOR_V3` appears at `:121`, `:131`, `:147`, plus the constructor loop at `:115` and `_executeSwap` branches. When a 7th ID is added, all need updating in lockstep. Consider an `_isValidProtocol(uint8)` internal pure helper.

### Tests

**`AetherExecutor.t.sol:1458-1501, 1503-1538` — pre-flashloan validation test is real, not a smoke test.**
```solidity
CountingAavePool countingPool = new CountingAavePool();
...
gatedExecutor.setDexEnabled(CURVE, false);
...
vm.expectRevert(abi.encodeWithSelector(AetherExecutor.ProtocolDisabled.selector, CURVE));
gatedExecutor.executeArb(...);
assertEq(countingPool.flashLoanCallCount(), 0, "flashloan must not fire when pre-check rejects");
```
Proves the revert fires **before** `flashLoanSimple`, not just that the outer call reverts. Also covers protocol=0 and protocol=7 bounds.

**`AetherExecutor.t.sol:1719-1756` — coinbase WETH fallback uses a real reverting-coinbase fixture.** Asserts (a) no ETH reaches coinbase, (b) WETH balance equals tip, (c) executor has no leftover.

**`AetherExecutor.t.sol:1655-1709` — UniV2 cap test is genuine.** Flashloan 500, over-spec amountIn 1000; asserts pool receives 500.

**🟡 MEDIUM — missing coverage:**
1. **No end-to-end test that `setDexRouter(BALANCER_V2, newVault)` actually routes subsequent swaps through the new vault.** This is **the central claim of the registry architecture** and isn't explicitly tested — `test_swapBalancer_vaultPattern` deploys via constructor, not via `setDexRouter`. Please add one.
2. **No test that Curve/Balancer/Bancor cap `forceApprove` at live balance** (because they don't — see MEDIUM above).
3. **No test that `setDexRouter(UNISWAP_V2, nonZeroAddr)` is a semantic no-op** (or revert on per-pool protocols). Currently accepts writes for any `1..=6`; runbook says "only valid for BALANCER_V2/BANCOR_V3" but code doesn't enforce.

**💡 NIT — protocol-constant sentinel test (`t.sol:1768-1781`) is circular.**
`assertTrue(executor.protocolEnabled(UNISWAP_V2))` uses a test-file constant also defined as `1`. Hardcode numeric IDs: `assertTrue(executor.protocolEnabled(1))` etc. The authoritative cross-language parity check lives Rust-side (PR2 scope).

### Docs

**`docs/incident-response.md` — correctly updates the ownership-rotation playbook** for Ownable2Step and documents the `transferOwnership(address(1))` footgun that would brick ownership under the new model.

**`docs/runbook.md` — new DEX workflow documents `setDexRouter` (router migration, no redeploy) and `setDexEnabled` (kill-switch)**, with the distinction between new-instance vs new-protocol-type clearly drawn.

**🔵 LOW — `setPaused` is not documented in the runbook.** Given `setPaused` is the on-chain circuit-breaker surface the Go risk manager should call, there should be a `cast send <EXECUTOR> "setPaused(bool)" true` example alongside the `ControlService/SetState` gRPC command. Operators need to know the overlapping-but-distinct semantics (Rust detection halt vs. on-chain `executeArb` halt).

**🔵 LOW — `docs/incident-response.md` alert-channels table references PagerDuty/Telegram/Discord.** Project convention is Slack-only; preexisting, but low-cost to fix here since this PR touches the file.

---

## Verdict — REQUEST CHANGES

Must-fix before merge:

1. **HIGH** — remove `gas: 2300` stipend on `block.coinbase.call` (`AetherExecutor.sol:244`). Real contract-coinbases will over-consume and silently fall through to WETH, losing inclusion priority.
2. **MEDIUM** — apply the live-balance `forceApprove` cap symmetrically to Curve/Balancer/Bancor (`:310-333`), or add explicit tests proving the on-chain revert is the intended behavior for those paths.
3. **MEDIUM** — file a follow-up issue for Timelock / multisig governance on `setDexRouter` / `setDexEnabled` before mainnet carries material TVL — these are the registry's substitute for `_authorizeUpgrade` and need equivalent governance.
4. **MEDIUM** — add an end-to-end test that `setDexRouter(BALANCER_V2, newVault)` routes subsequent swaps through the new vault.

Non-blocking follow-ups:

- Document `setPaused` in the runbook alongside `ControlService/SetState`.
- Fix PagerDuty/Telegram/Discord reference in `docs/incident-response.md` to Slack.
- Extract protocol-range check into `_isValidProtocol` helper.
- Harden the protocol-constant parity test with hardcoded numeric IDs.
- When PR2 lands, ensure its Curve/Balancer/Bancor calldata encoders respect the same live-balance cap as the on-chain approval (finding #2 has an engine-side half).

0xfandom (approved):
### Verdict — APPROVE

All four round-1 must-fix items are resolved or accepted with sound walk-back rationale. All four non-blocking items are addressed. The new routing test is a genuine proof of the registry's central architectural claim, not a smoke test.

### Per-item resolution

| # | Prior must-fix | Status | Evidence |
|---|---|---|---|
| 1 | HIGH — drop `gas: 2300` stipend on `block.coinbase.call` | Resolved | `AetherExecutor.sol:274` now reads `(bool sent,) = block.coinbase.call{value: tipAmount}("");` (no gas cap). WETH re-wrap fallback preserved at `:275-278` for the revert case. Commit `69a5df46`. |
| 2 | MEDIUM — live-balance `forceApprove` cap on Curve/Balancer/Bancor | Accepted with rationale | `_swapCurve:346-351`, `_swapBalancer:369-376`, `_swapBancor:385-392` call `forceApprove(..., step.amountIn)` unclipped. NatSpec at the top of each labels these "PULL-BASED PROTOCOL" and explains why a contract-side allowance cap only changes the revert error (from `insufficient balance` to `ERC20InsufficientAllowance`) without preventing the bundle loss — the pool reads the pull amount from `step.data`, not from the allowance. Fix correctly relocated engine-side in #97 ("…close poolId + live-balance gaps"). |
| 3 | MEDIUM — governance follow-up filed | Resolved | Issue #96 filed covering the timelock/multisig scope. Deferred intentionally — not implementing until mainnet TVL warrants it. |
| 4 | MEDIUM — e2e routing test `setDexRouter(BALANCER_V2, newVault)` | Resolved | `test_setDexRouter_balancerV2_routesToNewVault` at `AetherExecutor.t.sol:1417-1490`. Deploys two distinct `CountingBalancerVault` instances, executor constructed pointing at `firstVault`, calls `setDexRouter(BALANCER_V2, secondVault)`, executes real flashloan arb, asserts `secondVault.swapCallCount() == 1 && firstVault.swapCallCount() == 0` plus `allowance(..., secondVault) == 0` post-swap. Genuine routing + no-leakage proof. |

| # | Prior non-blocking | Status | Evidence |
|---|---|---|---|
| NB1 | Document `setPaused` in runbook | Resolved | `docs/runbook.md` new "Emergency Pause" section with `cast send … "setPaused(bool)" true` example, resume path, verification via `paused()(bool)`, and a table contrasting on-chain pause vs. Rust `ControlService/SetState`. |
| NB2 | Alert channels Slack-only | Resolved | `docs/incident-response.md:236-239` — PagerDuty/Telegram/Discord rows removed, single Slack row remains. |
| NB3 | Extract `_isValidProtocol` helper | Resolved | `AetherExecutor.sol:397-400` defines `internal pure` helper. Used at `setDexRouter:133`, `setDexEnabled:141`, and pre-flashloan validation in `executeArb`. Constructor loop at `:116-118` keeps explicit `for (uint8 p = UNISWAP_V2; p <= BANCOR_V3; p++)` bounds — init-time safety unchanged. |
| NB4 | Harden sentinel test with hardcoded IDs | Resolved | `test_protocolConstants_implicitlyMatch` uses `executor.protocolEnabled(1)` through `(6)` — hardcoded integers, anti-circular. |
| NB5 | PR2 engine-side live-balance cap | Tracked | Issue #97 scope confirmed to cover "live-balance gaps" alongside calldata encoders. Out-of-scope for this PR. |

### Regression scan — round-2 solidity delta

- `_isValidProtocol` is `internal pure` (correct — no storage read, inlineable). Used at all three runtime guards; constructor keeps explicit bounds.
- No modifier/auth changes on `setDexRouter`, `setDexEnabled`, or `executeArb` in the delta. `onlyOwner` intact on both admin functions.
- No new state variables. No events added or suppressed.
- Coinbase tip block at `:269-282`: on `!sent`, the path `IWETH(WETH).deposit{value: tipAmount}()` + `safeTransfer(block.coinbase, tipAmount)` replaces the old `CoinbaseTipFailed` revert. Note the behavior change: previously `!sent` reverted the flashloan; now the bundle succeeds with WETH tip. Matches acknowledged design intent — cost of "builders won't prioritize WETH tips" is paid to avoid bricking trades on contract-coinbase reverts.
- `_swapCurve` / `_swapBalancer` / `_swapBancor` retain vault/network derivation from `protocolRouter[...]` with `ZeroRouter` guard before use. Approval reset to 0 after each call preserved. Walk-back from `e313b77` is clean — no residual partial-cap code left behind.

### Regression scan — round-2 test delta

- `CountingBalancerVault` mock uses `transferFrom` driven by allowance (realistic pull pattern), increments `swapCallCount` per call. Assertion surface covers (a) migration target receives the call, (b) original target does NOT, (c) post-swap allowance is zeroed. Sufficient.
- Returning second hop is a `MockV2Pool` returning `flashAmount + premium + 10` — correctness depends on `MockV2Pool` (unchanged, not in this diff). Assumed stable.

### Evidence locations at current HEAD (`88773ec0` on `feat/executor-dex-registry`)

- `contracts/src/AetherExecutor.sol:116-118` — constructor loop
- `contracts/src/AetherExecutor.sol:132-137` — `setDexRouter`
- `contracts/src/AetherExecutor.sol:140-145` — `setDexEnabled`
- `contracts/src/AetherExecutor.sol:269-282` — coinbase tip + WETH fallback
- `contracts/src/AetherExecutor.sol:346-392` — `_swapCurve` / `_swapBalancer` / `_swapBancor`
- `contracts/src/AetherExecutor.sol:397-400` — `_isValidProtocol`
- `contracts/test/AetherExecutor.t.sol:1417-1490` — `test_setDexRouter_balancerV2_routesToNewVault`
- `contracts/test/AetherExecutor.t.sol:1864-1881` — hardened sentinel test
- `docs/runbook.md` "Emergency Pause" section
- `docs/incident-response.md:236-239` — Slack-only alerting table

PR #90, opened 2026-04-16
feat(engine): calldata builders + EXECUTOR_ADDRESS env (E4/WS-3 PR2)

Pablosinyores:
Stacked on #89. The engine half of #72.

This PR fills in the Rust/Go wiring needed to actually use the registry contract from #89: three previously-stubbed calldata builders, a compile-time Rust↔Solidity enum assertion, and mandatory EXECUTOR_ADDRESS env var loading on both services.

## Summary

- **`crates/simulator/src/calldata.rs`**: Curve `exchange`, Balancer V2 Vault `swap(SingleSwap, FundManagement, limit, deadline)` with GIVEN_IN, Bancor V3 `tradeBySourceAmount`. Each goes through `alloy::sol!`; selectors asserted via the generated `SELECTOR` constant rather than hardcoded bytes (ABI drift fails in the builder, not silently in a test).
- **`crates/common/src/types.rs`**: `const _: () = { assert!(ProtocolType::X as u8 == N); ... }` — reordering the enum or changing a discriminant now fails compilation, before any test runs. Also adds a doc comment on `EXECUTOR_OVERHEAD_GAS` explaining why the 30k absorbs the new registry SLOADs.
- **`crates/grpc-server/src/main.rs`**: new `parse_executor_address` fail-fast helper — rejects missing, unparseable, too-short, and zero-address values at startup.
- **`cmd/executor/main.go`**: same on the Go side via `parseExecutorAddressEnv`, using `common.IsHexAddress` + zero-check, stores canonical checksum hex in `Config.ExecutorAddr`. Error message redacts raw input (length only) to avoid leaking a pasted secret into `log.Fatalf` output.

## Dead-code status — builders have no callers yet (documented)

The three new builders (`build_curve_swap_calldata`, `build_balancer_swap_calldata`, `build_bancor_swap_calldata`) are library-only in this PR. `crates/grpc-server/src/engine.rs` still unconditionally constructs `SwapStep { ..., calldata: vec![] }` for every hop — every Curve/Balancer/Bancor hop ships empty `step.data` to the on-chain contract, which is absorbed by the post-swap `minAmountOut` revert. Not a regression (engine has always done this).

Wiring the builders into the call site — plus closing the Balancer `poolId` data-model gap and engine-side live-balance intersection — is tracked as **#97** and is out of scope here.

## Review-round-2 changes (since #90 first CHANGES_REQUESTED)

- **MEDIUM — Balancer selector test no longer self-referential.** Prior test compared two outputs from the same builder — if the builder's `sol!` signature were wrong, both encodings would share the wrong selector and the test would still pass. Fixed to declare a local canonical `sol!` swap signature and assert against `swapCall::SELECTOR.as_slice()`, matching the Curve/Bancor pattern. Test passing confirms the builder was correct all along.
- **HIGH — dead-code status documented** (this section).
- **LOW — Rust "too-short" env-var test added** for symmetry with Go (`parse_executor_address_too_short_is_error`).
- **LOW — Go error message redacts raw input.** `EXECUTOR_ADDRESS` parse error now reports length only; defends against a pasted-private-key footgun.
- Minimal `#[allow]` attributes added in `crates/simulator/src/fork.rs` and `crates/simulator/src/lib.rs` to keep `cargo clippy --tests -- -D warnings` green after the Balancer selector test change pulled those files into the gate. Logic unchanged; pre-existing conditions only.

## Follow-ups (not this PR)

- **#97** — wire the builders into `engine.rs`, add `pool_id: Option<[u8; 32]>` to `SwapStep` (Balancer needs the full 32-byte pool ID, not reconstructible from `Address` alone), intersect `amount_in` with live balance before encoding.
- **#96** — Timelock / multisig governance on the #89 registry setters before mainnet TVL.

## Test coverage

- 6 calldata tests (round-trip + selector stability via alloy-generated `SELECTOR` + field-layout asserts).
- 5 Rust env-var tests (unset / invalid hex / too-short / zero / valid).
- 7 Go env-var test-cases via table test (empty / whitespace / non-hex / too-short / zero / valid / whitespace-trimmed).

## Results

| Suite | Before | After |
|---|---|---|
| `cargo test -p aether-simulator` | 37 | **38** |
| `cargo test -p aether-common` | 22 | **22** + 1 compile-time assert block |
| `cargo test -p aether-grpc-server` | 88 | **89** |
| `go test -race ./cmd/executor/...` | green | **green** |

## Test plan

- [x] `cargo test -p aether-simulator -p aether-common -p aether-grpc-server` — all green
- [x] `go test -race ./cmd/executor/...` — all green
- [x] `cargo clippy -p aether-simulator -p aether-grpc-server --tests -- -D warnings` — clean
- [x] `go vet ./cmd/executor/...` — clean
- [ ] End-to-end against Anvil fork with #89 deployed executor (deferred to integration testing / #97)

## Notes for reviewer

- Base is `feat/executor-dex-registry` (#89), not `main` — this PR depends on the new ABI. Will rebase onto `main` after #89 lands.
- Curve indices stay as Rust primitive `i128` because alloy's `sol!` lowers Solidity `int128` to that type directly.

### DIFF
--- a/cmd/executor/main.go
+++ b/cmd/executor/main.go
@@ -12,6 +12,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 
 	"github.com/aether-arb/aether/internal/config"
@@ -73,7 +74,34 @@ func loadConfig() Config {
 		log.Printf("Config: GRPC_ADDRESS=%s (from env)", addr)
 	}
 
+	// EXECUTOR_ADDRESS is mandatory — without it bundles target the zero
+	// address and silently no-op on-chain. Parsed and validated here; any
+	// problem is a deployment misconfiguration, not a runtime condition.
+	if addr, err := parseExecutorAddressEnv(os.Getenv("EXECUTOR_ADDRESS")); err != nil {
+		log.Fatalf("Config: EXECUTOR_ADDRESS: %v", err)
+	} else {
+		cfg.ExecutorAddr = addr
+		log.Printf("Config: EXECUTOR_ADDRESS=%s (from env)", addr)
+	}
+
 	return cfg
+}
+
+// parseExecutorAddressEnv validates that the value is a non-zero, hex-parseable
+// Ethereum address. Returns a canonical (checksummed) hex string.
+func parseExecutorAddressEnv(raw string) (string, error) {
+	raw = strings.TrimSpace(raw)
+	if raw == "" {
+		return "", fmt.Errorf("unset — set to the deployed AetherExecutor contract address")
+	}
+	if !common.IsHexAddress(raw) {
+		return "", fmt.Errorf("%q is not a valid hex address", raw)
+	}
+	addr := common.HexToAddress(raw)
+	if addr == (common.Address{}) {
+		return "", fmt.Errorf("zero address is not a valid executor target")
+	}
+	return addr.Hex(), nil
 }
 
 // loadRiskConfig attempts to load risk parameters from config/risk.yaml,
--- a/cmd/executor/main.go
+++ b/cmd/executor/main.go
@@ -95,7 +95,7 @@ func parseExecutorAddressEnv(raw string) (string, error) {
 		return "", fmt.Errorf("unset — set to the deployed AetherExecutor contract address")
 	}
 	if !common.IsHexAddress(raw) {
-		return "", fmt.Errorf("%q is not a valid hex address", raw)
+		return "", fmt.Errorf("EXECUTOR_ADDRESS is not a valid hex address (input length=%d)", len(raw))
 	}
 	addr := common.HexToAddress(raw)
 	if addr == (common.Address{}) {
--- a/cmd/executor/main_test.go
+++ b/cmd/executor/main_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"math/big"
+	"strings"
 	"testing"
 	"time"
 
@@ -348,5 +349,39 @@ func TestSubmitter_CustomSubmitFn_IsUsed(t *testing.T) {
 	}
 	if results[0].Builder != "mock" {
 		t.Fatalf("expected builder mock, got %q", results[0].Builder)
+	}
+}
+
+func TestParseExecutorAddressEnv(t *testing.T) {
+	cases := []struct {
+		name    string
+		input   string
+		wantErr bool
+		wantLow string // lowercased checksum if no error
+	}{
+		{"empty is error", "", true, ""},
+		{"whitespace is error", "   ", true, ""},
+		{"non-hex is error", "not-an-address", true, ""},
+		{"too short is error", "0x1234", true, ""},
+		{"zero is error", "0x0000000000000000000000000000000000000000", true, ""},
+		{"valid lowercase parses", "0x1111111111111111111111111111111111111111", false, "0x1111111111111111111111111111111111111111"},
+		{"whitespace trimmed", "  0x2222222222222222222222222222222222222222  ", false, "0x2222222222222222222222222222222222222222"},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			got, err := parseExecutorAddressEnv(tc.input)
+			if tc.wantErr {
+				if err == nil {
+					t.Fatalf("expected error, got %q", got)
+				}
+				return
+			}
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if strings.ToLower(got) != tc.wantLow {
+				t.Fatalf("got %q, want (lowercased) %q", got, tc.wantLow)
+			}
+		})
 	}
 }
--- a/crates/common/src/types.rs
+++ b/crates/common/src/types.rs
@@ -1,7 +1,12 @@
 use alloy::primitives::{Address, U256};
 use serde::{Deserialize, Serialize};
 
-/// Protocol type enum matching on-chain constants in AetherExecutor.sol
+/// Protocol type enum matching on-chain constants in AetherExecutor.sol.
+///
+/// Discriminants are load-bearing: they are cast to `uint8` and encoded into
+/// `executeArb` calldata, where the on-chain contract branches on them. The
+/// const assertions below fail compilation if any discriminant drifts away
+/// from the Solidity constant it mirrors.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
 #[repr(u8)]
 pub enum ProtocolType {
@@ -12,6 +17,18 @@ pub enum ProtocolType {
     BalancerV2 = 5,
     BancorV3 = 6,
 }
+
+// Compile-time guard: if someone reorders the enum or changes a discriminant, the
+// build fails before a single test runs. Mirrors the Solidity constants in
+// `contracts/src/AetherExecutor.sol` (UNISWAP_V2..BANCOR_V3).
+const _: () = {
+    assert!(ProtocolType::UniswapV2 as u8 == 1);
+    assert!(ProtocolType::UniswapV3 as u8 == 2);
+    assert!(ProtocolType::SushiSwap as u8 == 3);
+    assert!(ProtocolType::Curve as u8 == 4);
+    assert!(ProtocolType::BalancerV2 as u8 == 5);
+    assert!(ProtocolType::BancorV3 as u8 == 6);
+};
 
 impl ProtocolType {
     /// Base gas cost for each protocol's swap
@@ -162,6 +179,11 @@ pub mod addresses {
 pub mod gas {
     pub const FLASHLOAN_BASE_GAS: u64 = 80_000;
     pub const TX_BASE_GAS: u64 = 21_000;
+    /// Catch-all overhead for `executeArb` framing: reentrancy guard, nonReentrant
+    /// modifier, calldata decoding, event emission, profit-split math, and per-step
+    /// dispatch. The 30k budget also absorbs the registry `SLOAD`s added in E4/WS-3
+    /// (2,100 gas cold on the first Balancer/Bancor hop of a block, ~100 gas warm on
+    /// subsequent reads) — worst case ~4,200 gas leaving >85% of the buffer unused.
     pub const EXECUTOR_OVERHEAD_GAS: u64 = 30_000;
     pub const UNIV3_PER_TICK_GAS: u64 = 5_000;
 }
--- a/crates/grpc-server/src/main.rs
+++ b/crates/grpc-server/src/main.rs
@@ -250,6 +250,17 @@ mod tests {
     }
 
     #[test]
+    fn parse_executor_address_too_short_is_error() {
+        // "0x1234" has the right prefix but is far fewer than 40 hex chars — it should
+        // fall into the invalid-hex bucket and fail, matching the Go side which already
+        // covers this input shape in its table test.
+        let _g = ENV_LOCK.lock().unwrap();
+        unsafe { std::env::set_var("EXECUTOR_ADDRESS", "0x1234"); }
+        assert!(parse_executor_address().is_err());
+        unsafe { std::env::remove_var("EXECUTOR_ADDRESS"); }
+    }
+
+    #[test]
     fn parse_executor_address_zero_is_error() {
         let _g = ENV_LOCK.lock().unwrap();
         unsafe {
--- a/crates/grpc-server/src/main.rs
+++ b/crates/grpc-server/src/main.rs
@@ -1,5 +1,7 @@
+use std::str::FromStr;
 use std::sync::Arc;
 
+use alloy::primitives::Address;
 use tokio::sync::RwLock;
 use tonic::transport::Server;
 use tracing::{error, info};
@@ -56,8 +58,14 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Create the AetherEngine with a broadcast sender connected to the
     // ArbService's stream.
     let arb_tx = arb_service.arb_sender();
+
+    // EXECUTOR_ADDRESS is mandatory — without it the engine builds calldata targeting
+    // the zero address, which is a silent no-op on-chain. Fail-fast at startup.
+    let executor_address = parse_executor_address()?;
+
     let engine_config = EngineConfig {
         rpc_url: std::env::var("ETH_RPC_URL").ok(),
+        executor_address,
         ..EngineConfig::default()
     };
     if engine_config.rpc_url.is_some() {
@@ -65,6 +73,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     } else {
         info!("ETH_RPC_URL not set — engine will use empty-state simulation");
     }
+    info!(executor_address = %engine_config.executor_address, "Executor contract target");
     let engine = Arc::new(AetherEngine::new_with_metrics(
         engine_config,
         arb_tx,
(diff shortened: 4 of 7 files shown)


0xfandom (changes requested):
# PR #90 Review — feat(engine): calldata builders + EXECUTOR_ADDRESS env (E4/WS-3 PR2)

Reviewed as PR2 of the Issue #72 split, stacked on #89. Only the delta `pr-89...pr-90` (405+/1-, 5 files) is reviewed here.

## Summary

This PR adds three swap-specific calldata builders (Curve `exchange`, Balancer V2 Vault `swap`, Bancor V3 `tradeBySourceAmount`), compile-time parity assertions for `ProtocolType` discriminants, and mandatory-on-startup `EXECUTOR_ADDRESS` parsing on both Rust and Go sides. Scope is tightly aligned with the title, the builders are clean, and the fail-fast patterns are well-formed. However, the new Curve/Balancer/Bancor builders are **not wired into any caller** in this PR — the engine still emits `calldata: vec![]` for every hop — so they currently ship as dead code. There is also a cross-PR correctness gap inherited from #89 that this PR could have closed but does not, plus a data-model gap for Balancer `poolId` that future wiring will need to address.

---

## Findings

### Calldata Correctness

**`crates/simulator/src/calldata.rs:127-136` — Curve `exchange` ABI — PASS (narrow scope).**
The 4-arg signature `exchange(int128 i, int128 j, uint256 dx, uint256 min_dy)` is correct for canonical Curve stableswap pools. `alloy::sol!` lowers `int128` to Rust `i128` directly, matching the comment. `step.pool.call(step.data)` in `AetherExecutor._swapCurve` (`contracts/src/AetherExecutor.sol:310-315`) does not read return data, so `returns (uint256)` is harmless. If `exchange_underlying` or 5-arg-with-receiver variants are intended later, a follow-up is needed, but the narrow scope here is fine.

**`crates/simulator/src/calldata.rs:145-197` — Balancer V2 Vault `swap` struct layout — PASS.**
Field order matches: `SingleSwap { poolId, kind, assetIn, assetOut, amount, userData }` and `FundManagement { sender, fromInternalBalance, recipient, toInternalBalance }`. `kind: 0` correctly encodes `GIVEN_IN`. `limit: min_amount_out` correctly applies slippage protection in the GIVEN_IN direction. Both `fromInternalBalance`/`toInternalBalance = false` is correct for executors that don't use internal balances.

**`crates/simulator/src/calldata.rs:203-225` — Bancor V3 `tradeBySourceAmount` signature — PASS.**
Arg count and order match BancorNetwork V3: `(sourceToken, targetToken, sourceAmount, minReturnAmount, deadline, beneficiary)`. `beneficiary = executor` is correct (`AetherExecutor.sol:328-333` routes the call and receives the output).

**🟠 HIGH — `crates/simulator/src/calldata.rs:127-225` — Dead code: the three new builders have no callers.**
Grep across the repo for `build_curve_swap_calldata|build_balancer_swap_calldata|build_bancor_swap_calldata` returns zero non-test hits. `crates/grpc-server/src/engine.rs:1046-1054` unconditionally constructs `SwapStep { ..., calldata: vec![] }` for every protocol. As of PR90, every Curve/Balancer/Bancor hop ships an empty `step.data` to the on-chain contract — `step.pool.call("")` / `balancerVault.call("")` / `bancorNetwork.call("")`. Not a regression (engine has always shipped empty calldata for these and the post-swap `minAmountOut` revert at `AetherExecutor.sol:261` catches it), but the PR's framing of "engine-side wiring" is only half-true: **library** is filled in, **call site** is not. Update the PR body or link a follow-up ticket.

**🟡 MEDIUM — `crates/simulator/src/calldata.rs:156` — Balancer builder takes `pool_id: [u8; 32]` but `SwapStep` only carries `pool_address: Address` (20 bytes).**
Balancer V2 `poolId` is 20-byte pool address left-padded to 32 bytes with a 12-byte specialization/nonce suffix — cannot be reconstructed from `Address` alone. `crates/common/src/types.rs:54-75` has no 32-byte field. The wiring PR will need either a new `SwapStep`/`ArbHop` field or a helper that reconstructs `poolId` from `(address, specialization)`. Flag now so the next PR is scoped correctly.

**🟡 MEDIUM — cross-PR: live-balance cap gap, inherited from #89.**
`_swapCurve`/`_swapBalancer`/`_swapBancor` (`AetherExecutor.sol:311, 319, 329`) all `forceApprove(step.amountIn)` without capping to live balance, unlike UniV2/Sushi branches. PR90's builders take `amount_in` verbatim with no live-balance plumbing. Not exploitable today because builders aren't called, but the follow-up wiring PR **must** either intersect with live balance before encoding, or the approval in the contract must be capped — otherwise any partial-fill dust will trigger `TRANSFER_FROM_FAILED` on every Curve/Balancer/Bancor hop.

**🔵 LOW — `crates/simulator/src/calldata.rs:127-225` — Deadline is a caller responsibility with no lower-bound defense.**
Builders accept `deadline: U256` and forward it verbatim. No assertion rejects `U256::ZERO` or `U256::MAX`. Current caller at `crates/grpc-server/src/engine.rs:1059` uses `U256::from(block_info.timestamp + 24)` (sane 2-block window), but a future caller could accidentally pass `U256::MAX` as a "no deadline" convenience. Consider `debug_assert!(deadline > 0 && deadline < U256::from(u64::MAX))` in each builder, or a shared `compute_deadline(block_ts: u64)` helper that's the only way to produce one.

### Rust/Solidity Parity

**`crates/common/src/types.rs:23-30` — Compile-time discriminant asserts — PASS.**
All six variants (`UniswapV2=1` through `BancorV3=6`) match `contracts/src/AetherExecutor.sol:27-32` exactly. Cast path (`ProtocolType as u8` → `protocol: uint8` in `SolSwapStep`) is exercised at `crates/simulator/src/calldata.rs:41`. Strictly better than runtime tests — fails before a single test runs.

**`crates/common/src/types.rs:163-170` — `EXECUTOR_OVERHEAD_GAS` doc — PASS.**
Explanation of 30k absorbing 2,100-gas cold `SLOAD` on first registry read per block is quantitatively correct per EIP-2929 (cold ≈ 2,100, warm ≈ 100).

### Executor-Address Wiring

**`crates/grpc-server/src/main.rs:218-228` — Rust `parse_executor_address` — PASS.**
Rejects missing, invalid hex, and zero address. Trims whitespace before parsing. Integrates cleanly with `?` at `main.rs:64`.

**`crates/grpc-server/src/main.rs:64-76` — Rust downstream wiring — PASS.**
Parsed `Address` placed on `EngineConfig { executor_address, ... }` at line 71, consumed at `crates/grpc-server/src/engine.rs:947`, passed to `EvmSimulator` for simulation target. End-to-end dataflow verified — not cosmetic.

**`cmd/executor/main.go:75-83` — Go fail-fast integration — PASS.**
`parseExecutorAddressEnv(os.Getenv(...))` inside `loadConfig`, wrapped in `log.Fatalf` on error. Note `loadConfig` no longer falls back to default `"0x0000..."` — the `defaultConfig()` value (`main.go:39`) is now effectively unreachable. Consider removing the default string or leaving a comment explaining it exists only for `defaultConfig()` unit-test purity.

**`cmd/executor/main.go:87-101` — Go `parseExecutorAddressEnv` — PASS.**
Rejects empty, whitespace-only (trimmed → empty), non-hex (via `common.IsHexAddress`), and zero. Returns EIP-55 checksummed `addr.Hex()`.
**🔵 LOW quibble:** on line 92, the error message wraps the raw input with `%q`. If a user pastes a private key by mistake into `EXECUTOR_ADDRESS`, it will end up in `log.Fatalf` output. Consider redacting to `len(raw)` or a fixed-length prefix.

**`cmd/executor/main.go:217, 262, 371` → `cmd/executor/bundle.go:54` — Go downstream wiring — PASS.**
`cfg.ExecutorAddr` flows into `consumeArbStream` → `processArb` → `bundler.BuildBundle(arb.Calldata, executorAddr, ...)`, then into the EIP-1559 `To:` field. Fail-fast is not cosmetic.

**🔵 LOW — `cmd/executor/bundle.go:54` — `common.HexToAddress` is silently lossy.**
On non-hex input, returns `Address{}` without error. Safe *today* because `loadConfig` guarantees validity, but the contract would be tighter if `Config.ExecutorAddr` were typed as `common.Address` rather than `string`. Easy future cleanup.

**Rust vs Go case normalization — PASS.** Both sides end up with EIP-55 checksummed strings. No equality-check risk observed.

### Tests

**`crates/simulator/src/calldata.rs:385-412` — Curve tests — PASS.**
Asserts selector against `exchangeCall::SELECTOR` (line 399) and exact calldata length `4 + 4*32` (line 400). `test_build_curve_swap_calldata_varies_with_inputs` catches swapped-index bugs.

**🟡 MEDIUM — `crates/simulator/src/calldata.rs:415-442` — Balancer selector test is self-referential.**
Does **not** compare against the alloy-generated `SELECTOR`. It only compares the first 4 bytes of two encodings produced by the same builder — if the `sol!` signature inside the builder were wrong (say, `SingleSwap` field order swapped), both encodings would share the wrong selector and the test would still pass. Asymmetric vs. Curve (line 399) and Bancor (line 478) — close the gap by declaring the same `function swap(SingleSwap,FundManagement,uint256,uint256)` in a local `sol!` in the test and asserting `swapCall::SELECTOR.as_slice()`.

**`crates/simulator/src/calldata.rs:467-497` — Bancor tests — PASS.**
Selector asserted against `tradeBySourceAmountCall::SELECTOR` (line 478), length `4 + 6*32` (line 481), beneficiary extraction at the last 32-byte word verified (line 497). Strong.

**Field-layout coverage — MEDIUM.**
No test hardcodes a known-good first-4-bytes byte literal. A single byte-level golden against the well-known `keccak256("exchange(int128,int128,uint256,uint256)")[:4]` / `swap((bytes32,uint8,address,address,uint256,bytes),(address,bool,address,bool),uint256,uint256)`[:4] / `tradeBySourceAmount(address,address,uint256,uint256,uint256,address)[:4]` would close the last mile. Two-line test per builder.

**`crates/grpc-server/src/main.rs:252-303` — Rust env-var tests — PASS with LOW gap.**
4 table cases (missing/invalid hex/zero/valid) cover the branches. Missing: "too short" case (e.g. `0x1234`). Rust's `Address::from_str` on `"0x1234"` hits the same "invalid hex" branch, so it's a coverage gap rather than a behavior gap. Go has an explicit case at `main_test.go:385`. Symmetry is cheap.

**`cmd/executor/main_test.go:379-410` — Go env-var tests — PASS.**
All 7 table cases distinct: empty, whitespace-only, non-hex, too-short, zero, valid-lowercase, whitespace-trimmed. `common.IsHexAddress` returns true only for exactly 42 chars (with `0x`) or 40 without, so "too short" is distinct from "non-hex" — former trips length check, latter trips hex-char check.

**Missing: end-to-end calldata→revm→executor test — deferred per PR body.** Accepted per project convention that Anvil end-to-end is deferred when explicit.

### Lints

**Pre-existing clippy lints not touched — PASS.**
PR body calls out "inconsistent digit grouping" and unused `setup_state_with_caller` as preexisting. Spot-checked: new code introduces no new clippy violations. `unsafe {}` env-var blocks are correct per Rust 2024. No `#[allow(clippy::...)]` silencing anything new.

### Stacked-PR Hygiene

**Base is `feat/executor-dex-registry` (#89), head is `pr-90` — PASS.**
Delta is cleanly additive, no reverts of #89. `EXECUTOR_OVERHEAD_GAS` doc update (`types.rs:163-170`) references #89's registry SLOADs — correct dependency direction.

---

## What's Good

- Compile-time `ProtocolType` discriminant asserts (`types.rs:23-30`) — strictly better than runtime tests.
- `alloy::sol!`'s generated `SELECTOR` used for Curve and Bancor selector assertions (the right pattern).
- Fail-fast on `EXECUTOR_ADDRESS` closes a real silent-misconfigure footgun — `bundler.BuildBundle` with `To: 0x00...` would no-op on-chain and the searcher would see "tx succeeded, zero profit" until someone noticed.
- Clear separation: builders are pure functions, env parsing isolated to startup.

---

## Verdict — REQUEST CHANGES

No critical-severity bugs, but two medium items should be addressed before merge, the dead-code scoping issue should at minimum be documented, and two cross-PR carryovers need tracking.

**Must-fix before merge:**

1. **HIGH** — Document the dead-code status. Update the PR body (or link a follow-up issue) stating that `build_curve_swap_calldata` / `build_balancer_swap_calldata` / `build_bancor_swap_calldata` have no callers yet and that `crates/grpc-server/src/engine.rs:1053` still ships `calldata: vec![]` for every Curve/Balancer/Bancor hop.
2. **MEDIUM** — Fix the Balancer selector test to assert against `swapCall::SELECTOR.as_slice()` instead of comparing two self-produced encodings. (`crates/simulator/src/calldata.rs:415-442`)

**Track as follow-up (blocking the wiring PR, not this one):**

3. **MEDIUM** — Cross-PR: live-balance cap. Wiring PR must intersect `step.amount_in` with live balance before encoding, or the on-chain `forceApprove(step.amountIn)` at `AetherExecutor.sol:311,319,329` must be capped. Either side works; both must not land without one.
4. **MEDIUM** — Balancer `poolId` data-model gap. Add a 32-byte field to `SwapStep`/`ArbHop` (or a helper) before `build_balancer_swap_calldata` can actually be called.

**Nice to have:**

5. Add a Rust "too short" env-var test for symmetry with Go.
6. Redact the raw input in the Go "not a valid hex address" error message (private-key-paste defense).
7. Add `debug_assert!` deadline sanity bounds to the three builders.
8. Type Go `Config.ExecutorAddr` as `common.Address` rather than `string`.

0xfandom (approved):
**Verdict: APPROVE.** All round-1 blockers resolved. Round-2 commits are surgical, no regressions. Security audit clean across 5 modified files.

---

### Round-1 item resolution table

| # | Round-1 Finding | Status | Evidence |
|---|---|---|---|
| 1 | HIGH — Document dead-code status of 3 builders | **Resolved** | PR body "Dead-code status" section + #97 confirmed with matching scope (wiring + poolId + live-balance). Grep confirms: engine only imports `build_execute_arb_calldata`; zero non-test callers of the three new builders. |
| 2 | MEDIUM — Balancer selector test self-referential | **Resolved** | `crates/simulator/src/calldata.rs:483-531` redeclares canonical `SingleSwap`/`FundManagement` + `swap(...)` in a local `sol!` block and asserts builder output against `swapCall::SELECTOR.as_slice()`. Genuinely cross-validating, not self-referential. Pattern now matches Curve (`exchangeCall::SELECTOR`) and Bancor (`tradeBySourceAmountCall::SELECTOR`). |
| 3 | MEDIUM — Cross-PR live-balance cap (Curve/Balancer/Bancor) | **Accepted — deferred to #97** | #97 scope covers it. Non-blocking for #90 since builders still unwired. |
| 4 | MEDIUM — Balancer `poolId` data-model gap | **Accepted — deferred to #97** | Same #97 coverage. |
| 5 | LOW — Rust "too short" env-var test | **Resolved** | Commit `9558b520` adds `parse_executor_address_too_short_is_error` with `"0x1234"`, proper `ENV_LOCK` guard, asserts `is_err()`. Parity with Go's too-short case. |
| 6 | LOW — Redact raw input in Go error | **Resolved** | Commit `b962dda3` replaces `%q` raw-wrap with `input length=%d` at `cmd/executor/main.go:98`. A 64-char private-key paste surfaces as `length=64` — enough signal for ops (0 / 42 / 64 distinguishable), no secret leak. No other log site echoes raw. |
| 7 | NIT — `debug_assert!` deadline bounds | Still open — accepted non-blocking |
| 8 | NIT — Go `Config.ExecutorAddr` as `common.Address` | Still open — accepted non-blocking |

---

### Calldata correctness — positive verification

- **Curve** (`calldata.rs:130-138`): Exact 4-arg canonical `exchange(int128, int128, uint256, uint256) returns (uint256)` — not `exchange_underlying`, not the 5-arg receiver variant. `i`/`j` typed as Rust `i128` mapping to Solidity `int128`. Test asserts `SELECTOR` + payload length `4 + 4*32 = 132`.
- **Balancer** (`calldata.rs:148-203`): Field order and types match V2 Vault ABI exactly — `SingleSwap{poolId, kind, assetIn, assetOut, amount, userData}`, `FundManagement{sender, fromInternalBalance, recipient, toInternalBalance}`. `kind = 0` (GIVEN_IN) hardcoded at `:187`; `limit = min_amount_out` at `:199` — correct GIVEN_IN slippage direction (limit is floor on output). `sender == recipient == executor`, both `internalBalance` flags `false`.
- **Bancor** (`calldata.rs:209-236`): Arg order matches V3 `tradeBySourceAmount(address, address, uint256, uint256, uint256, address)`. Test at `:590-605` asserts the last 32-byte word contains the beneficiary address right-aligned — catches any reordering.

### Enum parity — positive verification

`crates/common/src/types.rs:24-31` — `const _: () = { assert!(...) };` blocks at **module scope, NOT under `#[cfg(test)]`**. Fires in release builds. Solidity constants at `contracts/src/AetherExecutor.sol:29-34` map 1:1 (`UNISWAP_V2=1` through `BANCOR_V3=6`). Runtime asserts at `types.rs:207-214` under `#[cfg(test)]` are belt-and-suspenders redundancy.

### EXECUTOR_ADDRESS parsing — positive verification

- **Rust** (`grpc-server/src/main.rs:216-225`): missing → error, unparseable → error, zero → error. `raw.trim()` handles whitespace. 5 tests cover all branches.
- **Go** (`cmd/executor/main.go:92-105`): round-2 `length=%d` redaction confirmed at `:98`. `common.HexToAddress` returns EIP-55 checksummed via `.Hex()` at `:104`. `log.Fatalf` at `:81` embeds only the validated error string; no raw `os.Getenv` re-logged. 7 table-test cases cover empty / whitespace / non-hex / too-short / zero / valid / whitespace-trimmed.
- **Cross-language**: both sides case-insensitive downstream. No raw-string equality comparison in scope.

### `#[allow]` additions audit

- **`crates/simulator/src/fork.rs:328`** — `#[allow(clippy::unnecessary_map_or)]` on a single test assertion inside `#[cfg(test)]`. Pre-existing pattern, non-load-bearing.
- **`crates/simulator/src/lib.rs:332`** — `#[allow(dead_code)]` on test helper `setup_state_with_caller`. Test-only, round-1 explicitly noted this helper.

No new `unsafe`, no new `unwrap()` on user-input paths, no clippy silencing beyond these two.

### Regression scan

- Balancer test fix: new local `sol!` matches Balancer V2 Vault ABI exactly. Test passing ⇒ builder's internal `sol!` encodes identical signature ⇒ "builder was correct all along" claim holds.
- Go redaction preserves length=0/42/64 diagnostic signal.
- No working round-1 code was modified or regressed by the round-2 commits.

---

### Must-fix for this round

None.

### Non-blocking (tracked or accepted)

- #97 remains on the critical path before any caller can invoke the three builders (live-balance cap + `poolId` field).
- Nits 7 & 8 (deadline bounds, `common.Address` typing) remain accepted.

Ship it.

PR #94, opened 2026-04-20
fix(ingestion): decode Uniswap V2 / Sushi Swap events instead of dropping

### LINKED ISSUE #91: [E1/WS-1] Uniswap V2 / SushiSwap Swap events silently dropped by decode_log
## Context

`crates/ingestion/src/event_decoder.rs::event_topics()` (via `EventSignatures::swap_v2_topic()`) includes the Uniswap V2 / SushiSwap `Swap(address,uint256,uint256,uint256,uint256,address)` topic0 in the log subscription filter — so the RPC delivers these events to us.

But `decode_log` (`event_decoder.rs:101-126`) has **no branch** for `swap_v2_topic()`:

```rust
if topic0 == EventSignatures::sync_topic() {
    decode_sync(...)
} else if topic0 == EventSignatures::swap_v3_topic() {
    decode_swap_v3(...)
} else if topic0 == EventSignatures::token_exchange_topic() {
    decode_token_exchange(...)
} else if topic0 == EventSignatures::pair_created_topic() {
    decode_pair_created(...)
} else {
    None   // ← V2 Swap falls through here
}
```

So every V2 Swap on every monitored pool silently drops. Post PR #85, each drop increments `aether_decode_errors_total` — that's how the reviewer spotted it — but the counter rolls it up with genuine "unknown topic0" noise, so the signal isn't obvious without structured labels.

## Impact

On mainnet, every UniV2 and SushiSwap Swap log is invisible to Aether today. `decode_sync` catches the paired `Sync` event (V2 pools emit `Sync` on every state change), so reserves still reconcile and the price graph isn't stale — but any downstream logic that consumes `Swap` (per-trade slippage, volume, sender/recipient analytics) is flying blind for the two largest V2 protocols. Also: every monitored-pool V2 Swap currently contributes to `aether_decode_errors_total`, inflating the baseline and making real decode regressions harder to alert on.

## Scope

1. Implement `decode_swap_v2` in `crates/ingestion/src/event_decoder.rs`:
   - Parse `amount0In`, `amount1In`, `amount0Out`, `amount1Out` from `data` (4 × 32 bytes).
   - `sender` and `to` are indexed topics; include if useful downstream.
   - Decide the output `PoolEvent` shape: either extend `PoolEvent` with a dedicated `V2Swap` variant or map to an existing variant, consistent with how V3 Swap is modelled.
2. Add a branch to `decode_log` routing `swap_v2_topic()` to the new decoder.
3. Add a unit test analogous to `test_decode_v3_swap_event`: build a well-formed V2 Swap log, assert the decoder returns the correct variant.
4. Confirm no regression in `process_logs_decode_failure_increments_counter` (the test should still fire on a topic that really is unknown, not on a V2 Swap).

## Acceptance Criteria

- [ ] `decode_log` handles V2 Swap events end-to-end.
- [ ] Unit test decoding a V2 Swap from real event bytes.
- [ ] `aether_decode_errors_total` no longer bumps on V2 Swap events in discovery mode (verify in staging).
- [ ] `cargo test --workspace --release` + `cargo clippy -p aether-ingestion -p aether-grpc-server -- -D warnings` clean.

## Related

Discovered during PR #85 review. The observability PR exposed the silent drop via `aether_decode_errors_total`, but fixing the actual bug is out of that PR's scope.

Epoch: E1 (production correctness).

0xfandom:
## Summary
- `decode_log` had no branch for `swap_v2_topic()` — every UniV2 / Sushi Swap log silently returned `None`, dropping the event and inflating `aether_decode_errors_total`.
- Adds `PoolEvent::V2Swap` (informational, trade-level) alongside the existing `ReserveUpdate` which still drives state reconciliation via the paired `Sync` event.
- Adds `decode_swap_v2` parsing the two indexed address topics (`sender`, `to`) + the 4×32-byte data payload (`amount0In | amount1In | amount0Out | amount1Out`), routes it from `decode_log`, and handles the new variant in the engines `handle_pool_update` match so it stays exhaustive.

## Files Changed

| File | Why |
|------|-----|
| `crates/ingestion/src/event_decoder.rs` | New `PoolEvent::V2Swap` variant; new `decode_swap_v2` function; new else-if branch in `decode_log`; three new unit tests (happy path + insufficient topics + insufficient data). |
| `crates/grpc-server/src/engine.rs` | New `PoolEvent::V2Swap { .. }` arm in `handle_pool_update` — `debug!` no-op (reserves still flow through `Sync` → `ReserveUpdate`). Keeps the match exhaustive. |

## Acceptance Criteria

- [x] `decode_log` handles V2 Swap events end-to-end (`event_decoder.rs:115-116`)
- [x] Unit test decoding V2 Swap from real event bytes (`test_decode_v2_swap_event`)
- [x] `aether_decode_errors_total` no longer bumps on V2 Swap events — deterministic from code path (V2 Swap no longer hits the `None` else-arm). Staging verification out-of-band.
- [x] `cargo test --workspace --release` clean — **398 pass / 0 fail** (+3 new)
- [x] `cargo clippy -p aether-ingestion -p aether-grpc-server -- -D warnings` clean — **0 errors / 0 warnings**

## Design Note

`V3Update` and the new `V2Swap` now both follow the same pattern: dedicated variants for Swap events. The critical difference: V3 Swap carries pool *state* (`sqrtPriceX96`, `liquidity`, `tick`), so `V3Update` drives the price graph. V2 Swap carries only trade-level data (amounts + participants) — state reconciles via the paired `Sync` log on every V2 / Sushi pool interaction, which already drives `ReserveUpdate`. So `V2Swap` is informational: it exists so the event decodes (no decode error), so the match stays exhaustive, and so downstream trade analytics have a hook. No graph mutation in its engine arm.

## Test plan

- [x] `cargo test -p aether-ingestion --release` → 69 passed (was 66, +3 new V2 swap tests)
- [x] `cargo test -p aether-grpc-server --release` → 106 passed
- [x] `cargo test --workspace --release` → 398 passed, 0 failed
- [x] `cargo clippy -p aether-ingestion -p aether-grpc-server -- -D warnings` → clean
- [x] Spot-checked: existing `test_process_logs_unknown_event_ignored` still fires on truly-unknown topics (`B256::repeat_byte(0xFF)`); no regression.

Closes #91

### DIFF
--- a/crates/grpc-server/src/engine.rs
+++ b/crates/grpc-server/src/engine.rs
@@ -680,6 +680,26 @@ impl AetherEngine {
                     }
                 }
             }
+            PoolEvent::V2Swap {
+                pool,
+                sender,
+                to,
+                amount0_in,
+                amount1_in,
+                amount0_out,
+                amount1_out,
+            } => {
+                // Informational only — reserves reconcile via the paired
+                // `Sync` event, which arrives in the same log batch and
+                // drives `ReserveUpdate` above. This arm exists so the
+                // match stays exhaustive and downstream trade analytics
+                // have a hook.
+                debug!(
+                    %pool, %sender, %to,
+                    %amount0_in, %amount1_in, %amount0_out, %amount1_out,
+                    "V2 swap (informational)"
+                );
+            }
             PoolEvent::V3Update {
                 pool,
                 sqrt_price_x96,
--- a/crates/ingestion/src/event_decoder.rs
+++ b/crates/ingestion/src/event_decoder.rs
@@ -58,6 +58,18 @@ pub enum PoolEvent {
         reserve0: U256,
         reserve1: U256,
     },
+    /// V2 / Sushi Swap — informational (reserves still reconcile via the
+    /// paired `Sync` event; this variant exposes per-trade amounts and
+    /// participants for downstream analytics).
+    V2Swap {
+        pool: Address,
+        sender: Address,
+        to: Address,
+        amount0_in: U256,
+        amount1_in: U256,
+        amount0_out: U256,
+        amount1_out: U256,
+    },
     /// V3 state update
     V3Update {
         pool: Address,
@@ -114,6 +126,8 @@ pub fn decode_log(
 
     if topic0 == EventSignatures::sync_topic() {
         decode_sync(data, source_address, protocol_hint)
+    } else if topic0 == EventSignatures::swap_v2_topic() {
+        decode_swap_v2(topics, data, source_address)
     } else if topic0 == EventSignatures::swap_v3_topic() {
         decode_swap_v3(topics, data, source_address)
     } else if topic0 == EventSignatures::token_exchange_topic() {
@@ -123,6 +137,41 @@ pub fn decode_log(
     } else {
         None
     }
+}
+
+fn decode_swap_v2(topics: &[B256], data: &[u8], pool: Address) -> Option<PoolEvent> {
+    // V2 Swap(address indexed sender, uint256 amount0In, uint256 amount1In,
+    //        uint256 amount0Out, uint256 amount1Out, address indexed to)
+    //
+    // topics: [topic0, sender (indexed), to (indexed)]
+    // data:   4 × 32-byte words — amount0In | amount1In | amount0Out | amount1Out
+    if topics.len() < 3 || data.len() < 128 {
+        return None;
+    }
+
+    let sender = Address::from_slice(&topics[1].as_slice()[12..]);
+    let to = Address::from_slice(&topics[2].as_slice()[12..]);
+
+    let amount0_in = U256::from_be_slice(&data[0..32]);
+    let amount1_in = U256::from_be_slice(&data[32..64]);
+    let amount0_out = U256::from_be_slice(&data[64..96]);
+    let amount1_out = U256::from_be_slice(&data[96..128]);
+
+    trace!(
+        %pool, %sender, %to,
+        %amount0_in, %amount1_in, %amount0_out, %amount1_out,
+        "V2 Swap decoded"
+    );
+
+    Some(PoolEvent::V2Swap {
+        pool,
+        sender,
+        to,
+        amount0_in,
+        amount1_in,
+        amount0_out,
+        amount1_out,
+    })
 }
 
 fn decode_sync(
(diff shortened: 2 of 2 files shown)


0xfandom (commented):
## Summary

Adds a `V2Swap` informational variant and a `decode_swap_v2` branch to the event decoder so UniV2/Sushi Swap logs no longer fall through to the `None` arm and pollute `aether_decode_errors_total`. Deliberately does not mutate the price graph (reserves still reconcile via the paired `Sync`). Small, well-scoped, passes reasoning; ABI layout and address extraction are correct.

## Acceptance Criteria

| # | Criterion | Status | Evidence |
|---|---|---|---|
| 1 | V2 Swap event is decoded, not dropped | Met | `event_decoder.rs:127-128` routes `topic0 == swap_v2_topic()` into `decode_swap_v2`; happy-path test at `event_decoder.rs:367-406` exercises the full `decode_log` path and asserts `Some(V2Swap{..})`. |
| 2 | ABI layout matches on-chain `Swap(sender indexed, amount0In, amount1In, amount0Out, amount1Out, to indexed)` | Met | Solidity emits indexed params in declaration order: `topics[1]=sender`, `topics[2]=to`. Data is 4 consecutive U256 in declaration order. Decoder at `event_decoder.rs:146-152` matches exactly. Address extraction `&topics[N].as_slice()[12..]` is consistent with `decode_pair_created`. |
| 3 | `swap_v2_topic()` is in the RPC subscription filter | Met | `provider.rs event_topics()` includes `EventSignatures::swap_v2_topic()`; `test_event_topics_returns_known_signatures` asserts it's present. Logs actually arrive from the node. |
| 4 | Engine handles `V2Swap` variant | Met | `engine.rs:683-702` adds a `V2Swap` arm that emits `debug!` only. Match is exhaustive by variant (no wildcard) — without the arm it would fail to compile. Comment flags informational intent. |
| 5 | Existing `aether_decode_errors_total` counter still fires for genuinely unknown topics | Met | `test_process_logs_decode_failure_increments_counter` still exercises an all-0xFF unknown topic and asserts the counter advances. Not touched by this PR. |
| 6 | Negative tests for malformed input | Met | `event_decoder.rs:422-444` covers `topics.len() < 3` (missing `to`) and `data.len() < 128` (truncated amounts); both return `None`, matching the guard at `event_decoder.rs:142-144`. |
| 7 | No graph mutation (deliberate per issue #91) | Met | `engine.rs` arm is a single `debug!` — no `working_graph.lock()`, no `pool_registry` lookup. Reserves still reconcile via the `Sync` arm. |
| 8 | `aether_decode_errors_total` no longer bumps on V2 Swap in discovery mode (staging verification) | Deferred | PR explicitly defers to out-of-band staging verification. Decode path is unit-tested end-to-end, so the `None`-path that increments the counter is provably not taken. Worth eyeballing on merge-day metrics. |

## Blockers

None.

## Non-blocking observations

Triage: none cause deployment or post-deployment problems. All are future-direction nudges, not gates.

- **V2 swap counter / metric.** `engine.rs:683-702` emits `debug!` and nothing else. Issue #91's stated motivation included trade analytics (per-trade slippage, volume, sender/recipient). A `debug!` line isn't actionable at ops volume. An `aether_v2_swaps_decoded_total` counter would (a) give a positive-signal that the fix landed and (b) seed the analytics work. Follow-up ticket, not this PR.

- **Protocol is never identified.** `V2Swap` carries `pool: Address` but no `ProtocolType` — UniV2 vs SushiSwap is indistinguishable without a registry lookup. Couple with the counter work; same follow-up.

- **Log-level split.** Decoder uses `trace!`, engine uses `debug!`. Each consistent with its file's local convention. Only matters if `RUST_LOG=debug` in prod, which shouldn't happen; even then the sibling V2 reserve-update arm already does the same, so no net new volume.

- **No engine-arm test.** Decoder tests cover decoding; the arm is a single `debug!` with no failure mode. Worth 10 lines once real logic (counter, metric) lands.

- **`V2Swap` variant size (~224 B).** `PoolEvent` is cloned across `tokio::sync::broadcast` — but broadcast uses `Arc` internally, so the cost is ref-count bumps, not memcpy. Negligible. Worth noting only if the enum grows further.

## Verdict

**APPROVE** — ship it.

Tight, correctly-scoped fix. ABI decoding is provably correct (Solidity indexed-param ordering + consistent with sibling decoders), the subscription filter already receives these logs, the existing decode-failure counter test still guards the regression surface, and the engine arm is a deliberate no-op consistent with the issue's "reserves reconcile via Sync" invariant. The one meaningful gap — a V2-swap-specific metric — is appropriately punted since this PR is fixing the drop-on-decode bug, not building the analytics pipeline. Staging verification of the error-counter deflation is an acceptable out-of-band check.

Pablosinyores (approved):
## Review — APPROVE

Tight, minimal fix. V2 Swap events were being dropped because `decode_log` had no routing branch; this PR adds the variant, the decoder, and the exhaustive engine arm.

### Correctness
- ABI layout matches `Swap(address indexed sender, uint256 amount0In, uint256 amount1In, uint256 amount0Out, uint256 amount1Out, address indexed to)` — topics `[sig, sender, to]` + 128-byte data payload in declaration order. Decoder matches exactly.
- Bounds guard (`topics.len() < 3 || data.len() < 128`) returns `None` gracefully on malformed logs. No panics.
- `U256::from_be_slice` on fixed slices and `Address::from_slice` on `&[u8; 20]` are both infallible here.
- Design is sound: V2 `Swap` is trade-level only; pool state still reconciles via the paired `Sync` event → `ReserveUpdate`. The `debug!`-only engine arm is intentional — no graph mutation.

### Security
- No secrets in diff.
- No hot-path heap allocation beyond the enum variant itself.
- Graceful handling of untrusted input.

### Tests
- Happy path with realistic USDC/ETH values, plus insufficient-topics and insufficient-data error paths.
- Existing `test_process_logs_unknown_event_ignored` still passes — no regression on unknown topics.
- `cargo test --workspace --release` → 398 pass; `cargo clippy -D warnings` clean.

### Suggestions (non-blocking, follow-up)
- **SUG1**: Consider a `aether_v2_swaps_decoded_total` counter. The engine arm is `debug!`-only, so confirming the fix landed in prod requires indirect verification via the decrement in `aether_decode_errors_total`. A counter is a cheap one-liner.
- **SUG2**: `PoolEvent::V2Swap` has no `ProtocolType` discriminator, so UniV2 vs Sushi is indistinguishable without a registry lookup. Not needed today, but will matter once trade analytics consume the variant.

Commit message is conventional and clean. Ready to merge.

PR #93, opened 2026-04-17
feat(shadow): live mainnet shadow-mode runner + WS transport preference

0xfandom:
## Summary
- Orchestrates the **full production pipeline** (aether-rust + aether-executor) against live Ethereum mainnet, with `AETHER_SHADOW=1` gating the executor off Flashbots. Every `ARB PUBLISHED` is dumped as JSON for post-run forensics; script exits non-zero if any bundle actually reaches a builder.
- **Patches `ProviderConfig::default()`** to prefer `ETH_WS_URL` over `ETH_RPC_URL` so the engine picks WebSocket push subscriptions over HTTP polling when both are set.
- Validated with two back-to-back live mainnet runs; caught a real +0.027 ETH arb on UniswapV3 stablecoin pairs.

## Files Changed

| File | Why |
|------|-----|
| `crates/grpc-server/src/provider.rs` | One-line fix: read `ETH_WS_URL` first, fall back to `ETH_RPC_URL`. Without this, even with both env vars set, the provider picked HTTP → 1s polling latency per block. |
| `scripts/shadow_mode_live.sh` | New 5-phase orchestrator: preflight → build → start Rust (live WS) → start executor (shadow) → 30s metric poll loop → final snapshot. Wraps in `caffeinate -i` to stop macOS idle-sleep from stalling mid-run. Hard-exits non-zero if any bundle is actually submitted. |

## Acceptance Criteria

- [x] Runs against live mainnet WS (no Anvil, no local fork)
- [x] Executor bundles are built + signed but NEVER sent — `aether_executor_bundles_submitted_total` stays at 0
- [x] Per-arb JSON forensics (`reports/shadow_<ts>/bundles/<arb_id>.json`) with path, hops, flashloan, raw_tx_hex, net_profit
- [x] `transport=WebSocket` confirmed in Rust log (patch working)
- [x] No laptop-sleep stall (caffeinate active, 900s requested = 900s wall-clock)
- [x] Clean exit — 0 ERROR/WARN over 15-min run

## Live Mainnet Validation

Two back-to-back runs this session:

| Metric | 10-min run | 15-min run |
|---|---|---|
| Blocks processed | 50 | **75** (perfect 12s mainnet cadence) |
| Cycles detected | 927 (auto-discovery of new pool jumped the count mid-run) | 0 (quiet window on 21-pool majors) |
| Simulations run | 3 | 0 |
| Shadow bundles built | 3 (same `arb_id` × 3) | 0 |
[...]

PR #115, opened 2026-04-28
feat(ledger): trade-ledger foundation — schema + no-op access layer

### LINKED ISSUE #95: [E2] Trade ledger: Postgres schema + access layer (Rust + Go)
## Context

CLAUDE.md specifies Postgres as the trade ledger (line 23, 263) and Redis as the state cache — **neither exists in code**. Zero DB client dependencies, zero migrations, zero `.sql` files, no `postgres` service in `deploy/docker/docker-compose.yml`.

This issue builds the Postgres layer as a **first-class production persistence tier** for the whole bot. The layer must work identically in live, shadow, and historical-replay modes.

### What we can't answer today

- "How many arbs has the bot detected this week?"
- "What's our win rate by pool path, by builder, by hour of day?"
- "Has detection latency regressed since commit X?"
- "For the bundle that reverted on mainnet, what did we simulate vs what landed?"
- "What's our realized P&L — gas spent minus profit — per day?"
- "Of the arbs mev-inspect shows landed on mainnet in the last 24h, which ones did we also detect?"

Every one of these is a trivial SQL query against the schema below, impossible today.

### What it unlocks

- **Queryable history** instead of resettable Prometheus counters
- **Regression detection** across commits (diff two runs on latency / catch rate)
- **Unblocks #74** (MEV strategy bandit requires persistent `(features → reward)` training data)
- **Real catch-rate numbers** by joining against ground-truth mev-inspect data
- **P&L accounting ledger** — the thing CLAUDE.md actually means by "trade ledger"
- **Post-mortem forensics** — one SQL query instead of `grep` across log directories
- **SQL-backed Grafana dashboards** — already have Grafana in docker-compose

### 1. Infrastructure

- Add `postgres:17` service to `deploy/docker/docker-compose.yml` with named volume + healthcheck
- `DATABASE_URL` env var (libpq format). Binaries must be **no-op backward-compatible** when unset so local dev without Docker still works.

### 2. Schema (initial migration `migrations/0001_trade_ledger.sql`)

Minimum viable ledger:

- **`arbs`** — every opportunity the engine publishes. PK = `arb_id`. Columns cover detection context (block, path, hops, protocols, pool addresses), sizing (flashloan token + amounts), economics (gross / net profit wei, gas estimate, tip bps), timing (detection_us, sim_us), and provenance (git_sha).

- **`bundles`** — every bundle built (shadow or live). FK to `arbs`. Captures signed tx hex, gas used, shadow flag, submitted timestamp, fan-out builder list.

- **`inclusion_results`** — one row per (bundle, builder) pair. Populated when `GetBundleStats` resolves. Captures inclusion, landed block, landed tx hash, error.

- **`pool_registry`** — the merged view of configured + auto-discovered pools the bot is currently monitoring. First-seen timestamp, protocol, token pair, tier.

- **`pnl_daily`** — rollup: per-day realized profit wei, gas spent wei, bundle count, inclusion count. Populated by a scheduled job once per UTC day.

All timestamps `TIMESTAMPTZ`. U256 values stored as `NUMERIC(78,0)`. Variable-shape data (path, protocols, pool_addresses) as `JSONB` so queries can index into them.

Indexes on: `arbs(target_block)`, `arbs(ts DESC)`, `arbs(path_hash)`, `bundles(arb_id)`, `inclusion_results(bundle_id)`, `inclusion_results(included_block)`.

### 3. Access layer

- **Rust** — `crates/common/src/db.rs` wrapping `sqlx::PgPool`. Methods: `insert_arb`, `insert_pool`, `update_inclusion`, plus typed row structs.
- **Go** — `internal/db/ledger.go` wrapping `pgxpool`. Methods: `InsertBundle`, `InsertInclusion`.
- Both sides: connect from `DATABASE_URL`; if unset, return a no-op `Ledger` that logs once and becomes a silent sink.

### 4. Wiring into production paths

- Rust engine: on every `ARB PUBLISHED`, also `ledger.insert_arb(...)`. On auto-registered pool, also `ledger.insert_pool(...)`.
- Go executor: on bundle build (shadow or live), `ledger.InsertBundle(...)`. On `GetBundleStats` poll resolving, `ledger.InsertInclusion(...)`.
- Risk manager / monitor: read-only reporters; no writes from these processes.

### 5. Migration tooling

- `sqlx-cli` as workspace dev-dependency
- `migrations/` directory at repo root
- `scripts/db_migrate.sh` loads `.env`, runs `sqlx migrate run`
- CI spins up `postgres:17` as a service container for the Rust integration-test job

### 6. Docs

- `.env.example` adds `DATABASE_URL=postgres://aether:aether@localhost:5432/aether`
- CLAUDE.md updated: the two Postgres lines now reference the actual schema path + env var

## Acceptance Criteria

- [ ] `docker compose up postgres` brings up Postgres 17, healthcheck passing
- [ ] `scripts/db_migrate.sh` applies `0001_trade_ledger.sql` idempotently
- [ ] With `DATABASE_URL` set: a short live / shadow / replay run writes rows to `arbs`, `bundles`, `pool_registry` and `inclusion_results` (for the real-submission path) as events occur
- [ ] With `DATABASE_URL` unset: both binaries boot and run identically to today (no writes, no errors)
- [ ] Row counts match corresponding Prometheus counters for the same time window (e.g. `SELECT count(*) FROM arbs WHERE ts > $WINDOW` == `aether_arbs_published_total` delta)
- [ ] `cargo test --workspace --release` + clippy clean; `go test ./... -race` passes
- [ ] CI integration-tests job runs against a real Postgres service container

## Out of Scope — follow-up issues

- **Redis state cache** (CLAUDE.md's second storage tier) — separate hot-path concern, different access pattern
- **mev-inspect-py importer** populating a `landed_arbs` ground-truth table
- **Grafana dashboards** querying this schema
- **Retention / partitioning** via `pg_partman` once we have meaningful data volume
- **Read replicas / `pgbouncer`** — scale-out concern
- **Backfill** of any historical JSON dumps into the schema — can be a one-off importer script later

## Related

- CLAUDE.md lines 23, 263 (architecture intent)
- Unblocks #74 (MEV strategy bandit — needs training data)
- Prerequisite for real catch-rate / P&L dashboards in the observability epoch

## Epoch

E2 — production observability foundation. Unblocks live-mode analytics and every downstream data-driven feature.

0xfandom:
> **Heads-up for reviewers — this PR does NOT close #95.**
> It ships only the **foundation slice** of the trade-ledger work: schema + no-op access layer. The Postgres-backed writes, engine / executor wiring, scheduled rollup, and CI integration-test container all land in **separate follow-up PRs** (links in *Out of scope* below). #95 stays open until those land.
>
> Why split: the full issue touches Rust + Go + infra + CI + docs (~25-40 files) and has merge-conflict surface with in-flight PRs (#93 shadow, #98 obs, #114 polling). Foundation-only keeps this PR mergeable today with **zero behaviour change** when `DATABASE_URL` is unset.

## Summary

- Adds `postgres:17` compose service (healthcheck + named volume).
- Adds `migrations/0001_trade_ledger.sql` — 5 tables: `arbs`, `bundles`, `inclusion_results`, `pool_registry`, `pnl_daily`. U256 → `NUMERIC(78,0)`, variable-shape path data → `JSONB`, all timestamps `TIMESTAMPTZ`. Indexes per the issue spec.
- Adds `scripts/db_migrate.sh` — sqlx-cli wrapper, loads `.env`, applies migrations.
- Adds `Ledger` trait + `NoopLedger` default in **Rust** (`crates/common/src/db.rs`) with `NewArb` / `NewPool` / `InclusionUpdate` payloads.
- Adds `Ledger` interface + `NoopLedger` default in **Go** (`internal/db/ledger.go`) with `NewBundle` / `NewInclusion` / `PnLDailyDelta` payloads.
- `.env.example` documents `DATABASE_URL` + `POSTGRES_USER/PASSWORD/DB` (commented out so unset behaviour stays the default).

**Backward compatibility:** binaries still build and run identically to current `main`. `NoopLedger` discards every write and logs once at startup so operators can grep for `ledger disabled`.

## Files Changed

| File | Purpose |
|---|---|
| `migrations/0001_trade_ledger.sql` | Initial schema (5 tables + indexes) |
| `scripts/db_migrate.sh` | sqlx-cli migrate wrapper |
| `crates/common/src/db.rs` | Rust `Ledger` trait + `NoopLedger` |
| `crates/common/src/lib.rs` | Export `db` module |
| `crates/common/Cargo.toml` | Promote `serde_json` + `tracing` to runtime deps |
| `internal/db/ledger.go` | Go `Ledger` interface + `NoopLedger` |
| `internal/db/ledger_test.go` | Go ledger smoke test |
| `deploy/docker/docker-compose.yml` | `postgres:17` service + `postgres-data` volume |
| `.env.example` | `DATABASE_URL` + `POSTGRES_*` lines |

## Acceptance Criteria (issue scope satisfied by this PR)

- [x] `docker compose up postgres` brings up Postgres 17 with healthcheck (compose config validated; reviewer to do live `up`)
- [x] `scripts/db_migrate.sh` is in place; applies `0001_trade_ledger.sql` idempotently against any reachable Postgres
- [x] With `DATABASE_URL` unset: both binaries boot and run identically to today (no writes, no errors)
- [x] `cargo build --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p aether-common` clean
- [x] `go build ./...`, `go test ./... -race -count=1`, `go vet ./...` clean
- [ ] **Deferred to follow-up PRs:** rows actually written by live / shadow / replay; row counts matching Prometheus counters; CI integration-tests job against a real Postgres service container

## Test plan

- [x] `cargo build --workspace` — green
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo test -p aether-common --release` — 40 pass incl new NoopLedger tests
- [x] `go build ./...` — green
- [x] `go test ./... -count=1` — green incl new ledger smoke test
- [x] `go vet ./...` — clean
- [x] `docker compose -f deploy/docker/docker-compose.yml config` — postgres service + `postgres-data` volume registered correctly
- [ ] Reviewer manual: `docker compose up postgres -d` + `DATABASE_URL=postgres://aether:aether@localhost:5432/aether ./scripts/db_migrate.sh` applies the migration

## Out of scope — tracked under #95, will land in follow-ups

- **Rust `PgLedger`** — `sqlx::PgPool` impl of `Ledger`; engine wiring (`insert_arb` on every `ARB PUBLISHED`, `insert_pool` on auto-registered pools); workspace `sqlx` + `uuid` deps.
- **Go `PgLedger`** — `pgxpool` impl; executor wiring (`InsertBundle` on bundle build, `InsertInclusion` on `GetBundleStats` resolve).
- **`pnl_daily` rollup** — scheduled job, once per UTC day.
- **CI** — `postgres:17` service container in the Rust integration-test job, schema-applied + round-trip test.
- **Redis state cache** — separate hot-path concern, different access pattern.
- **mev-inspect-py importer + Grafana dashboards backed by this schema** — downstream of wiring.

Refs #95


Pablosinyores (changes requested):
## Review — feat(ledger): trade-ledger foundation

Foundation slice is well-scoped and the "zero behaviour change when `DATABASE_URL` unset" promise holds in code. Verified locally: `cargo build -p aether-common`, `cargo test -p aether-common --lib db::` (2/2), `go build ./internal/db/...`, `go test ./internal/db/...`, `go vet ./internal/db/...`, `docker compose config`, `bash -n scripts/db_migrate.sh` — all clean. Schema↔payload mapping walked column-by-column on both sides.

Hot-path latency impact verified zero: `NoopLedger` is empty-body vtable dispatch, no alloc, no atomic. `OnceLock`/`sync.Once` only fires in the constructor.

### Blocking (CRITICAL)

**1. `internal/db/ledger.go:40` — `Error string` is not nullable**

SQL `inclusion_results.error TEXT` is nullable. Go zero-value `""` will write an empty string instead of NULL, so `WHERE error IS NULL` returns zero rows for successful inclusions. Rust counterpart at `crates/common/src/db.rs:70` correctly uses `Option<String>`.

```diff
-    Error         string
+    Error         *string
```

**2. `internal/db/ledger.go:39` — `LandedTxHash []byte` has no length contract**

Schema is semantically a 32-byte tx hash; Rust counterpart at `crates/common/src/db.rs:69` is `Option<B256>` (exactly 32 bytes). The unbounded `[]byte` lets a caller silently write 0/20/64 bytes — Rust readers would then panic or get garbage. Suggest `*[32]byte` (nullable) or at minimum a documented length contract.

- **`go.mod:29`** — `github.com/google/uuid` is marked `// indirect` but is now imported directly in `internal/db/ledger.go:16`. Any `go mod tidy` will rewrite `go.mod`/`go.sum`. Run `go mod tidy` and commit the result.
- **`crates/common/src/db.rs:117-136`** — `uuid_compat::Uuid` has no documented byte-order contract. `google/uuid` is RFC 4122; if the future `uuid::Uuid` swap uses a different layout, the FK correlation `arbs.arb_id` ↔ `bundles.arb_id` breaks silently. Add `// TODO: replace with uuid::Uuid (RFC 4122 byte order) before PgLedger impl ships` on the struct so the swap is gated on the byte-layout match.
- **`crates/common/src/db.rs:52`** — `NewPool.protocol: String` is stringly-typed; `ProtocolType` enum already exists at `crates/common/src/types.rs:7`. Casing drift in the DB will silently break `WHERE protocol = ...` filters. Bind through the enum's `Display` impl.
- **`deploy/docker/docker-compose.yml:177`** — postgres service starts unconditionally on every `docker compose up`, consuming port 5432 and ~50MB RAM even when no binary is wired to it. The `reth` service at line 199 uses `profiles: [node]` for the same reason. Add `profiles: [ledger]` (or similar) to preserve the "no surprise" contract for current operators.
- **Clock-authority inconsistency** — `internal/db/ledger.go:25` (`SubmittedAt`) and `:41` (`ResolvedAt`) are client-set `time.Time`s that override the SQL `DEFAULT now()`, while `arbs.ts` has no Rust field and is DB-authoritative. If the Go executor's clock drifts vs Postgres, `submitted_at`/`resolved_at` aren't safely comparable with `arbs.ts`. Either document this as intentional (Go clock = "when Go acted", DB clock = "when row was written") or unify.

### Nice to Have (SUGGESTION)

- **`migrations/0001_trade_ledger.sql:62`** — `inclusion_results_bundle_id_idx` is redundant with the composite PK `(bundle_id, builder)` (line 59); the PK B-tree already serves any leading-column query. Drop it.
- **`internal/db/ledger.go:57-59`** — interface methods take structs by value. `NewBundle.SignedTxHex` is a ~500B string. NoopLedger is fine (compiler elides), but the future PgLedger is forced to copy on every call. Pointer receivers cost nothing now and avoid a future interface-breaking change.
- **`crates/common/src/db.rs:81`** — trait methods are infallible by design. This boxes the future PgLedger into log-and-drop, which works for fire-and-forget but rules out caller-side retry/metric emission. Consider `-> Result<(), LedgerError>` with an extension trait that auto-logs-and-drops for hot-path callers.

### What's Good

- Single coherent commit, conventional format, clean message.
- 9 files, no drive-by edits, scope matches the PR body.
- `NUMERIC(78,0)` correctly sized for U256 (2^256 has 78 digits). `BYTEA`/`TIMESTAMPTZ` choices are correct.
- `OnceLock<()>` / `sync.Once` startup-warn is symmetric across both binaries with identical log message, so operators can grep `ledger disabled` on either side.
- Trait object safety is explicitly tested (`db.rs:151`).
- No existing code imports `aether_common::db` or `internal/db` outside the new files — runtime behaviour vs `main` is genuinely unchanged.

### Verdict

**REQUEST CHANGES** — gated on the two CRITICAL items above. Once those land, the WARNINGs are cheap fixes that lock the schema/trait into a shape that won't fight the PgLedger follow-up.

Pablosinyores (approved):
## Re-review — all prior findings addressed

Verified locally on `e4fc873`: `cargo build -p aether-common`, `cargo test -p aether-common --lib db::` (2/2), `cargo clippy -p aether-common --all-targets -- -D warnings`, `go build ./internal/db/...`, `go test ./internal/db/...`, `go vet ./internal/db/...` — all clean. CI green (rust 2m55s, go 29s, solidity 17s, toolchain 4s). `mergeStateStatus: CLEAN`.

### Blocking findings — both resolved

**`internal/db/ledger.go:48-49`** — `LandedTxHash *[32]byte` and `Error *string` now match the Rust counterpart's nullability + length contract. The struct doc explicitly calls out the bugs the previous shapes would have caused (`""` instead of NULL; arbitrary-length tx hashes), which closes the loop nicely for a future reader.

### Non-blocking findings — all resolved

- **`go.mod:7-12`** — `github.com/google/uuid` and the four `go.opentelemetry.io/otel` packages promoted to the direct require block. `go mod tidy` is now a no-op for contributors.
- **`crates/common/src/db.rs:124-137`** — `uuid_compat::Uuid` byte-order contract documented as RFC 4122 network byte order, with explicit notes on Postgres binary wire format compatibility, `uuid::Uuid::as_bytes`/`from_bytes` round-trip, and the gotcha around `from_bytes_le` / Windows GUIDs. The future `uuid::Uuid` swap is now a typed contract, not an implicit assumption.
- **`crates/common/src/db.rs:59`** — `NewPool.protocol` is now `ProtocolType`, not `String`. Stringly-typed casing drift in the DB is no longer possible. `crates/common/src/types.rs` also derives `Default` mapping to `ProtocolType::UniswapV2 = 1` so the `Default` derive on `NewPool` keeps working — the rationale is in the doc comment, which I think is the right safe sentinel.
- **`deploy/docker/docker-compose.yml`** — `postgres` service is now gated behind `profiles: [ledger]`. Verified: `docker compose config --services` returns 9 services (no postgres); `docker compose --profile ledger config --services` returns 10 (postgres added). The "no surprise" contract for current operators is intact.
- **`migrations/0001_trade_ledger.sql:11-22`** — clock-authority policy is now an explicit comment in the schema header: event-time columns (`arbs.ts`, `bundles.submitted_at`, `inclusion_results.resolved_at`) are CLIENT-SET, persistence-boundary columns (`pool_registry.first_seen/last_seen`, `pnl_daily.updated_at`) are DB-SET. The reasoning (event-vs-row time skew under load) is captured. Future wiring PRs have a clear contract.

### One minor follow-up (non-blocking, just for tracking)

The new clock-authority policy says `arbs.ts` is CLIENT-SET, but `NewArb` at `crates/common/src/db.rs:30-48` still has no `ts` field. The wiring follow-up that ships `PgLedger` will need to add `pub ts: chrono::DateTime<Utc>` (or equivalent) to `NewArb` and have callers populate it at arb-creation time. Same applies to `NewBundle.SubmittedAt` and `NewInclusion.ResolvedAt` on the Go side, which already have client-set fields — those are good. Just calling out the asymmetry so it doesn't slip when `PgLedger` lands.

Quality of the fixes is high — author didn't just patch the field types but also added doc comments explaining the bugs the original shapes would have caused, which makes the intent durable.

**Verdict: APPROVE**

PR #122, opened 2026-05-05
feat(ledger): Go PgLedger pgxpool impl + executor wiring (PR-2 of #95)

0xfandom:
## Summary

PR-2 of the trade-ledger plan. Adds the Go side of the writer surface: `PgLedger` on `pgxpool`, executor wiring on bundle submission and per-builder result handling, inline `pnl_daily` roll-up, and `aether_ledger_*` metrics that mirror the Rust side exactly. After this lands, running aether against a fork or staging produces row trails in `bundles`, `inclusion_results`, and `pnl_daily` — closing the practical "DB integration on develop" goal.

PR-3 (CI Postgres + counter-vs-row reconciliation test) is intentionally **not** in scope per the agreed-to minimum-viable plan.

## Single commit

`39492a2` feat(ledger): Go PgLedger pgxpool impl + executor wiring

## What landed

**`internal/db/ledger_pg.go`** (new, 232 LOC)
- `PgLedger` on `pgxpool`. Hot-path enqueue is non-blocking `try-send` onto a bounded `chan` (cap 1024). Dispatcher goroutine drains and fans out via a `Semaphore(8)` so the pgx pool runs at capacity without acquire-queueing.
- Saturation drops the row and bumps `aether_ledger_drops_total{op}` — never blocks the executor.
- 2 s connect timeout: misconfigured `DATABASE_URL` fails fast and `LedgerFromEnv` falls back to `NoopLedger`.
- Idempotent inserts on `bundles` + ON CONFLICT upsert on `inclusion_results` + delta-accumulation upsert on `pnl_daily` (multiple writers can contribute to the same day without lost updates; `NUMERIC(78,0)` preserves U256 economics losslessly).
- `Close()` for graceful shutdown drains in-flight writes.

**`internal/db/metrics.go`** (new, 51 LOC)
- `LedgerMetrics` registered against the default Prometheus registry. Names byte-identical to the Rust side:
  - `aether_ledger_writes_total{op, result}`
  - `aether_ledger_drops_total{op}`
  - `aether_ledger_queue_depth`
  - `aether_ledger_write_latency_ms{op}`

**`internal/db/ledger.go`** (extended)
- `ArbIDNamespace` + `BundleIDNamespace` constants. `ArbIDNamespace` is **byte-identical** to the Rust `ARB_ID_NAMESPACE` so log↔DB join keys are symmetric across the gRPC boundary; a `TestArbIDNamespaceMatchesRust` pin guards drift.
- `ArbIDFromOppID(string) → uuid.UUID` and `BundleIDFor(uuid, block) → uuid.UUID` for deterministic ids. Same `(arb, block)` always produces the same `bundle_id` so resubmission `ON CONFLICT` is naturally idempotent.

**`cmd/executor/main.go`** (wired)
- `LedgerFromEnv` at startup, deferred `Close()` on shutdown.
- `processArb` gains `db.Ledger` param. After bundle build, derives `arb_db_id` + `bundle_id` deterministically and emits both as structured log fields on the `arb submitted` line so `grep <id> logs/* | psql ... WHERE arb_id = ...` works straight from logs.
- Shadow path persists a `bundles` row with `IsShadow=true` (no builders).
- Live path persists `bundles` + per-builder `inclusion_results`. Critical semantic note (in code comment): `Included` here reflects builder *acceptance* for next-block inclusion, **not** on-chain inclusion — the future `GetBundleStats` poll loop will UPSERT the same `(bundle_id, builder)` row with `included_block` + `landed_tx_hash` populated.
- `pnl_daily` inline roll-up: `bundle_count` bumps every submit; on builder acceptance `realized_profit_wei` + `inclusion_count` accumulate. ~30 LOC to keep the table populated during fork / staging runs without a separate cron.

**Tests**
- All 14 `processArb` + 2 `consumeArbStream` call sites in `*_test.go` updated to pass `db.NewNoopLedger()`.
- New deterministic-id tests pin `ArbIDNamespace` byte-equality with Rust and the determinism / variance contract on `ArbIDFromOppID` / `BundleIDFor`.
- `go test ./...` 100% green.

**Cargo / go.mod**
- Adds `github.com/jackc/pgx/v5/pgxpool`.

## Acceptance criteria progress (#95)

| # | Criterion | After this PR |
|---|-----------|----|
| 5 | Rust access layer (`sqlx::PgPool`) | ✅ (PR #119) |
| **6** | **Go access layer (`pgxpool`)** | ✅ |

Criteria 8 (counter-vs-row reconciliation) and 9 (CI Postgres) intentionally deferred per the minimum-viable plan.

## Out of scope

- `GetBundleStats` poll loop that UPSERTs `inclusion_results` with on-chain truth. Schema already supports the upsert; this PR's writes are submission-time. Followup.
- CI Postgres service container + integration tests + counter-vs-row reconciliation. Tracked as future PR-3.
- Chain-backfill reconciliation worker (drives the dead-code Rust `update_inclusion`).
- Retention / partitioning on `arbs` + `inclusion_results`.

## Local verification

- `go build ./...` clean.
- `go vet ./...` clean.
- `go test ./...`: all green (executor + db + risk + config + pooldiscovery).

### DIFF
--- a/cmd/executor/bench_test.go
+++ b/cmd/executor/bench_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aether-arb/aether/internal/db"
 	"github.com/aether-arb/aether/internal/risk"
 	"github.com/aether-arb/aether/internal/testutil"
 )
@@ -76,7 +77,7 @@ func BenchmarkProcessArb(b *testing.B) {
 	b.ResetTimer()
 	b.ReportAllocs()
 	for i := 0; i < b.N; i++ {
-		_, _ = processArb(ctx, arb, time.Now(), rm, bundler, submitter,
+		_, _ = processArb(ctx, arb, time.Now(), rm, bundler, submitter, db.NewNoopLedger(),
 			"0x0000000000000000000000000000000000000000", 0.5)
 	}
 }
--- a/cmd/executor/integration_test.go
+++ b/cmd/executor/integration_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aether-arb/aether/internal/db"
 	aethergrpc "github.com/aether-arb/aether/internal/grpc"
 	pb "github.com/aether-arb/aether/internal/pb"
 	"github.com/aether-arb/aether/internal/risk"
@@ -75,7 +76,7 @@ func TestProcessArbViaGRPC(t *testing.T) {
 	}
 
 	rm, bundler, submitter := newTestComponents()
-	submitted, err := processArb(ctx, arb, time.Now(), rm, bundler, submitter,
+	submitted, err := processArb(ctx, arb, time.Now(), rm, bundler, submitter, db.NewNoopLedger(),
 		"0x0000000000000000000000000000000000000000", 0.5)
 	if err != nil {
 		t.Fatalf("processArb: %v", err)
@@ -114,7 +115,7 @@ func TestConsumeArbStream(t *testing.T) {
 
 	lb := NewLiveBalance()
 	lb.Set(0.5)
-	consumeArbStream(ctx, client, bundler, submitter, rm,
+	consumeArbStream(ctx, client, bundler, submitter, rm, db.NewNoopLedger(),
 		"0x0000000000000000000000000000000000000000", lb)
 
 	// Verify bundle tracking was updated
@@ -132,7 +133,7 @@ func TestCircuitBreakerAcrossArbs(t *testing.T) {
 
 	// Process first arb — should succeed
 	arb1 := testutil.ProfitableTriangleArb()
-	submitted, err := processArb(ctx, arb1, time.Now(), rm, bundler, submitter,
+	submitted, err := processArb(ctx, arb1, time.Now(), rm, bundler, submitter, db.NewNoopLedger(),
 		"0x0000000000000000000000000000000000000000", 0.5)
 	if err != nil {
 		t.Fatalf("arb1: %v", err)
@@ -150,7 +151,7 @@ func TestCircuitBreakerAcrossArbs(t *testing.T) {
 
 	// Process second arb — should be rejected by risk manager
 	arb2 := testutil.Profitable2HopArb()
-	submitted, err = processArb(ctx, arb2, time.Now(), rm, bundler, submitter,
+	submitted, err = processArb(ctx, arb2, time.Now(), rm, bundler, submitter, db.NewNoopLedger(),
 		"0x0000000000000000000000000000000000000000", 0.5)
 	if err != nil {
 		t.Fatalf("arb2: %v", err)
@@ -183,7 +184,7 @@ func TestMixedArbScenarios(t *testing.T) {
 			ctx := context.Background()
 
 			arb := tc.arb()
-			submitted, err := processArb(ctx, arb, time.Now(), rm, bundler, submitter,
+			submitted, err := processArb(ctx, arb, time.Now(), rm, bundler, submitter, db.NewNoopLedger(),
 				"0x0000000000000000000000000000000000000000", tc.ethBalance)
 			if err != nil {
 				t.Fatalf("processArb: %v", err)
@@ -222,7 +223,7 @@ func TestGracefulShutdown(t *testing.T) {
 	go func() {
 		lb := NewLiveBalance()
 		lb.Set(0.5)
-		consumeArbStream(ctx, client, bundler, submitter, rm,
+		consumeArbStream(ctx, client, bundler, submitter, rm, db.NewNoopLedger(),
 			"0x0000000000000000000000000000000000000000", lb)
 		close(done)
 	}()
--- a/cmd/executor/main.go
+++ b/cmd/executor/main.go
@@ -23,6 +23,7 @@ import (
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/aether-arb/aether/internal/config"
+	"github.com/aether-arb/aether/internal/db"
 	aethergrpc "github.com/aether-arb/aether/internal/grpc"
 	pb "github.com/aether-arb/aether/internal/pb"
 	"github.com/aether-arb/aether/internal/risk"
@@ -212,6 +213,18 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	// Trade ledger: graceful no-op when DATABASE_URL is unset, so dev /
+	// shadow runs against a fork keep working without Postgres. Connect
+	// failure also degrades to NoopLedger so a misconfigured URL does not
+	// stall executor boot.
+	ledgerMetrics := db.NewLedgerMetrics()
+	ledger := db.LedgerFromEnv(ctx, os.Getenv("DATABASE_URL"), ledgerMetrics)
+	defer func() {
+		if pg, ok := ledger.(*db.PgLedger); ok {
+			pg.Close()
+		}
+	}()
+
 	// Initialize components
 	nonceManager := NewNonceManager(0)
 	if txSigner != nil {
@@ -299,7 +312,7 @@ func main() {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			consumeArbStream(ctx, grpcClient, bundler, submitter, riskMgr, execCfg.ExecutorAddress, liveBalance)
+			consumeArbStream(ctx, grpcClient, bundler, submitter, riskMgr, ledger, execCfg.ExecutorAddress, liveBalance)
 		}()
 	}
 
@@ -327,6 +340,7 @@ func processArb(
 	rm *risk.RiskManager,
 	bundler *BundleConstructor,
 	submitter *Submitter,
+	ledger db.Ledger,
 	executorAddr string,
 	ethBalance float64,
 ) (submitted bool, err error) {
@@ -372,6 +386,16 @@ func processArb(
 	}
 	buildSpan.End()
 
+	// Derive deterministic ledger ids before either branch so the same
+	// (arb, target_block) pair always maps to the same row, regardless of
+	// shadow vs live submission. Mirrors the Rust engine's
+	// `arb_id_for_opp` so log↔DB joins work end-to-end across the gRPC
+	// boundary.
+	arbDBID := db.ArbIDFromOppID(arb.Id)
+	targetBlock := arb.BlockNumber + 1
+	bundleID := db.BundleIDFor(arbDBID, targetBlock)
+	signedTxHex := signedTxsHex(bundle)
+
 	// Shadow mode: the bundle is fully built and signed, but we skip the
 	// network submission. Used by historical replay + pre-prod measurement
 	// to exercise the full pipeline without touching Flashbots.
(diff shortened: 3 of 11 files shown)


0xfandom:
## Self-review: alignment vs #95 + correctness on what landed

Reviewed locally before requesting external eyes. Posting here so the rationale travels with the PR.

### Alignment matrix (#95 acceptance criteria)

| # | Criterion | Status |
|---|---|---|
| 1 | `docker compose up postgres` healthcheck | N/A — pre-existing |
| 2 | `db_migrate.sh` idempotent | N/A — pre-existing |
| 3 | `DATABASE_URL` set → rows in `arbs`/`bundles`/`inclusion_results` | **Partial / broken** (see CRIT 1 + 2) |
| 4 | `DATABASE_URL` unset → identical boot | Met (`ledger_pg.go:275-288`, `main.go:220-226`) |
| 5 | Rust sqlx | Landed in #119 |
| 6 | **Go pgxpool** | Met (`ledger_pg.go:46-288`) — surface correct, behaviour buggy |
| 7 | Wiring | Met (`main.go:315, 421-507`) |
| 8 | Row counts ↔ Prom counters | Deferred → PR-3 |
| 9 | CI Postgres + `-race` integration | Deferred → PR-3 |

### Justified deferrals — and their consequences

- **Criteria 8 + 9 → PR-3**: justified per the agreed minimum-viable plan, but the consequence is severe — the exact integration test being deferred would have caught CRIT 1 and CRIT 2 below on first contact with a real Postgres. Deferring reconciliation while merging an unverified writer is the worst-of-both outcome.
- **`GetBundleStats` poll loop deferred**: justified — schema accommodates the future UPSERT. Consequence: `inclusion_results.included` here means *builder acceptance for next-block inclusion*, not on-chain landing. The partial index `inclusion_results_included_block_idx WHERE included` (`migrations/0001_trade_ledger.sql:76`) was clearly designed for on-chain truth, so any dashboard built between PR-2 and PR-3 will mis-report landed counts. Acknowledged in the PR comment at `main.go:450-454`; flagging the consequence, not the choice.
- **Chain-backfill / `update_inclusion` Rust dead code deferred**: justified per #95 out-of-scope. Consequence: `included_block` / `landed_tx_hash` stay NULL until that worker exists.
- **Retention / partitioning deferred**: justified for PR-2. `bundles.signed_tx_hex` is unbounded TEXT — at ~200 bundles/min × multi-tx, crosses 10 GB inside two months. Track it.

### CRITICAL — block merge

**CRIT 1 — `internal/db/ledger_pg.go:191-202` — JSONB column bound as `text[]`.** `bundles.builders` is `JSONB NOT NULL` (`migrations/0001_trade_ledger.sql:58`) but the Go insert binds `b.Builders` (`[]string`) directly. pgx v5 encodes `[]string` as Postgres `text[]`. Every live-mode insert will fail with `column "builders" is of type jsonb but expression is of type text[]`. Shadow mode dodges this only because shadow passes `Builders: nil` (`main.go:428`). **Fix**: `json.Marshal(b.Builders)` and bind `[]byte`, or change the column to `TEXT[]`. Pick one.

**CRIT 2 — `cmd/executor/main.go:421, 460` vs `migrations/0001_trade_ledger.sql:52` — FK race against `arbs`.** `bundles.arb_id` is `NOT NULL REFERENCES arbs(arb_id)`. Rust writes `arbs` via its own async channel; Go writes `bundles` via a separate async channel. There is no happens-before between them. Failure modes: (a) **shadow runs without a Rust ledger writer** → every `InsertBundle` violates FK; (b) **live under load** → the Go writer can drain a bundle row before the Rust writer drains the corresponding arb row → intermittent FK violations surfacing only as `aether_ledger_writes_total{op="insert_bundle",result="err"}`. **Fix options**: insert the `arbs` row from the Go side too with `ON CONFLICT (arb_id) DO NOTHING` (the executor has all the data on the gRPC `ValidatedArb`), or drop the FK and rely on application-level join discipline. Former is cleaner.

### HIGH

**H1 — `internal/db/ledger_pg.go:100-104` — `Close()` does not actually drain.** Order in `cmd/executor/main.go` is: `cancel()` → `wg.Wait()` (executor goroutines) → deferred `pg.Close()`. By the time `Close` runs, the outer ctx is already cancelled, and `runOne` (`:158-183`) uses that same ctx — so every drained op fails with `context.Canceled`. The "drains in-flight writes" doc comment at `:98-99` is wrong. **Fix**: give the writer pool an independent ctx with a hard deadline (e.g. 5s) used inside `runOne`, separate from the dispatcher's cancellation signal.

**H2 — `internal/db/ledger_pg.go:217-232` — UPSERT clobbers future poll results on resubmit.** The `ON CONFLICT (bundle_id, builder) DO UPDATE SET included = EXCLUDED.included, included_block = EXCLUDED.included_block, ...` overwrites a previously poll-populated `(included=true, included_block=N, landed_tx_hash=H)` row with a fresh submission carrying `included=false, NULL, NULL`. `BundleIDFor(arb, target_block)` is deterministic so retries collide by design. **Fix**: only UPDATE columns that are non-NULL in EXCLUDED, or move submission-time writes to a `submission_attempts` table and let the poll be the sole writer of `inclusion_results`.

**H3 — `cmd/executor/main.go:469-482` — `included=true` on builder acceptance pollutes the on-chain semantic.** Schema's partial index makes the intent unambiguous. Until PR-3's poll lands, dashboards filtering `WHERE included` will count accepted-but-not-landed bundles as landed. **Recommendation**: write `included = false` initially and let the poll flip it, or add a separate `accepted` boolean now.

### MEDIUM

**M1 — `internal/db/ledger_pg.go:138-156` — dispatcher leaks on ctx cancel.** When `ctx.Done()` fires, the loop returns with the channel still open — future enqueuers fall to the default branch and bump drops (fine), but goroutines already past the `sem` acquire are not awaited; the deferred `inflight.Wait()` is unreachable from that exit path. **Fix**: restructure so the loop breaks to a label and `inflight.Wait()` is always reached.

**M2 — `internal/db/ledger_pg.go:120-131` — `QueueDepth` gauge is racy.** `Inc()` runs after the send, `Dec()` runs before dispatch — under contention the gauge can transiently go negative. Cosmetic, but the Rust side likely uses the same metric name and cross-process scrapes will look weird. Use a periodic `Set(len(l.ch))` instead.

**M3 — `internal/db/ledger_pg.go:235-258` — `pnl_daily.bundle_count` inflates on resubmits.** Postgres serialises the row lock so the upsert itself is correct, but every submission bumps the counter — even when the deterministic bundle row was deduped in `bundles` via `DO NOTHING`. **Fix**: rebuild `pnl_daily` from a periodic `INSERT … SELECT COUNT(*)`, or only bump when the bundle INSERT actually inserted (use `RETURNING` to detect).

**M4 — `internal/db/ledger_test.go:23-34`** — namespace pin is correctly byte-identical to the Rust constant at `crates/grpc-server/src/engine.rs:182-185`. ✓ Gap: `BundleIDNamespace` has no Rust mirror so no equivalent pin is possible — add an explicit comment in `ledger.go` near the constant saying so, to future-proof against a Rust mirror drifting silently.

### LOW / NIT

- `internal/db/metrics.go:43-47` — histogram buckets start at 0.5ms; p50 of a single-row INSERT on local Postgres is ~150-300µs. Add 0.1, 0.25 buckets.
- `internal/db/ledger_pg.go:158-183` — `runOne` tags `context.Canceled` as `result="err"`. Distinguish `result="cancelled"` so shutdowns don't pollute the err-rate alert.
- `cmd/executor/main.go:497-507` — `gasSpentApprox` uses float64 to compute a value bound for `NUMERIC(78,0)`. The column exists precisely so we don't drop wei. Compute in `*big.Int` from the start.
- `internal/db/ledger.go:1-7` — package doc still says "a pgxpool-backed implementation … land in a follow-up". Stale now that the impl ships in this PR.

### What's already right

- Connect-failure-degrades-to-Noop matches the Rust contract; `LedgerFromEnv` is the right shape (`ledger_pg.go:275-288`).
- `ArbIDFromOppID` is byte-pinned to the Rust namespace with a meaningful failure message.
- Bounded channel + dispatcher + 8-way semaphore is the right structural pattern; capacity / pool size / inflight are tuned in lockstep with a clear explanatory comment.

### Verdict

**Request changes.** Architecture is right. Criterion 6 nominally met, but criterion 3 is effectively unmet — CRIT 1 kills every live `bundles` insert and CRIT 2 makes shadow inserts impossible without a Rust writer. The deferral of PR-3 integration tests is the direct enabler of both bugs slipping through. Fix CRIT 1 + 2 + H1 (`Close()` drain), then re-evaluate H2/H3 (or accept them as PR-3 work with explicit tracking issues).

0xfandom:
Self-review round 1 addressed across 5 atomic commits (`39492a2..93981c3`). Triage notes — for each finding, marked **fixed** / **deferred** / **noise**.

**Fixed (5 commits)**

- `8a08dae` fix(ledger): marshal `NewBundle.builders` to JSONB bytes — **CRIT 1**. `[]string` → `text[]`; column is `JSONB`. Every live insert would have failed. Now `json.Marshal` + explicit `$8::jsonb` cast.
- `15944e8` fix(ledger): drop `bundles.arb_id` FK via migration `0002_drop_bundles_arb_id_fk.sql` — **CRIT 2**. Rust + Go writers race; FK fired immediately on insert. Migration drops the constraint with a comment spelling out the trade-off (orphan bundles surface as `aether_ledger_writes_total{op="insert_arb",result="err"}` and are informative on LEFT JOIN).
- `c300451` perf(ledger): bounded `Close()` drain + guaranteed dispatcher cleanup — **CRIT 3 + MED 1**. `Close()` now caps drain at 5 s, cancels the dispatcher's independent context on timeout. `dispatch()` uses `defer inflight.Wait()` so every exit path drains. Dispatcher runs on a context independent of the caller's so cancelling ctx no longer kills in-flight queries.
- `92a764d` fix(ledger): submit-time `Included=false`; defer on-chain truth to poll loop — **HIGH 2**. `inclusion_results.included` is the on-chain outcome the schema's `WHERE included` partial index expects, not builder ACK. Submit-time row writes `Included=false` with builder error preserved; future `GetBundleStats` poll UPSERTs the on-chain truth. Drops the included-branch from `pnl_daily` for the same reason.
- `93981c3` chore(ledger): queue gauge race + sub-ms buckets + bigInt gas + namespace doc — **MED 2 + MED 4 + 2 NITs**. `enqueue()` Inc-before-send + revert-on-fail; histogram gains 0.1 / 0.25 ms buckets; `gasSpentApprox` is now bigInt-only (no float intermediate) so cumulative `pnl_daily` doesn't drift; `BundleIDNamespace` documents why no Rust pin test exists today.

**Deferred (real but follow-up scope)**

- **HIGH 1** UPSERT overwrites poll-populated columns on resubmit — only matters once `GetBundleStats` poll loop ships. No poll loop in this PR; resubmit-after-poll is a hypothetical race for now. Tracked.
- **MED 3** `pnl_daily.bundle_count` inflates on `processArb` retries — would need pgx `RowsAffected()` plumbing through the async path or a separate counter rebuild. Small impact (< few % drift under retry). Tracked.
- **NIT** Tag context-cancel as separate metric reason — defer.

**Noise (skipped)**

- "Package doc still claims follow-up impl" — already accurate enough for the trait + ships-with-impl status.

**Local checks**

- `go build ./...` clean.
- `go vet ./...` clean.
- `go test ./...` all green (executor + db + risk + config + pooldiscovery + monitor).

Re-requesting review.

PR #127, opened 2026-05-07
feat(pools,engine): V3 / Curve / Balancer post-state predictors + state cache

0xfandom:
Closes #124 (foundation half — see *Out of scope* below).

Lands the analytical post-state predictors and the live pool-state cache the mempool decode pipeline needs to call them. After this merges, every protocol family the engine tracks has a `predict_post_state` method on its pool struct and a corresponding `PoolState` cache entry kept in lockstep with the price graph.

## What ships

| Layer | Item |
|---|---|
| Predictor math | `UniswapV3Pool::predict_post_state` (single-tick + tick-crossing detection) |
| Predictor math | `CurvePool::predict_post_state` (2-coin StableSwap, Newton get_y) |
| Predictor math | `BalancerPool::predict_post_state` (weighted-2-token, equal- and unequal-weight branches) |
| State cache | `PoolState` enum + `PoolStateCache` (`Arc<DashMap<Address, Arc<PoolState>>>`) + factory |
| Engine wire-up | V2 / Sushi cache populated at bootstrap via `getReserves()` + on `PoolEvent::ReserveUpdate` |
| Engine wire-up | V3 cache populated at bootstrap via `slot0()` + on `PoolEvent::V3Update` |
| Engine wire-up | Curve cache populated at bootstrap via `A() + balances(0) + balances(1)` |
| Engine wire-up | Balancer V2 cache populated at bootstrap via `getPoolId() → vault.getPoolTokens(poolId) → pool.getNormalizedWeights()` |
| Dispatch | `predict_post_state_with_fallback` wrapper returns `Option<UnifiedPostState>` and invokes a caller-supplied closure on low-confidence paths |
| Observability | `aether_sim_evm_fallback_total{reason}` Prometheus counter with stable label set |

## Confidence flags

Each predictor surfaces a single boolean signal that the wrapper uses to escalate to an EVM fork-replay fallback:

| Pool family | Field | When it clears |
|---|---|---|
| V3 | `single_tick` | swap moved `sqrt_price_x96` past the active tick bucket |
| Curve | `analytical` | Newton iteration produced an invalid post-state |
| Balancer | `analytical` | unequal-weight pool — first-order Taylor approximation deemed too coarse |

The wrapper bumps `aether_sim_evm_fallback_total{reason}` with one of `v3_tick_crossed` / `curve_unconverged` / `balancer_unequal_weight` / `unknown_protocol` and returns `None`, so dashboards show how often the analytical path is insufficient.

## PoolMetadata change

`PoolMetadata` grows a `tick_spacing: Option<i32>` field so the V3 cache can build a valid `UniswapV3Pool` (which needs tick_spacing to know where the active tick bucket ends). New `register_pool_with_tick_spacing` constructor accepts the value; the original `register_pool` delegates with `None` so existing callers compile unchanged. Bootstrap reads the value from `pools.toml` `tick_spacing` field; non-V3 pools store `None`.

## Tests

```
pools crate                  55 unit tests   ✓
grpc-server bin              50+ unit tests  ✓
workspace cargo test         all green        ✓
```

New tests cover:
- V3 / Curve / Balancer predictor parity vs the existing `get_amount_out` math (regression guard so the sim and the detector never drift)
- Single-tick vs cross-tick V3 fixtures with the bucket boundary engineered to either keep the swap inside or push it out
- Cache round-trip (variant + state field accessibility) for every pool family
- Wrapper dispatch (escalates on cleared confidence, returns state on high confidence)

## Out of scope (deferred to follow-up PRs)

- **Real EVM fork-replay implementation behind the fallback.** Today's `EvmSimulator` is built around `executeArb` calldata + the AetherExecutor contract; a generic "apply this transaction and read the affected pool's post-state" entry point is its own substantial change. The metric in this PR tells us how often the fallback is needed; the implementation lands when that infra catches up.
- **Anvil parity tests against historical mainnet swaps.** The unit tests above cover predictor / cache / dispatch correctness; a fixture set asserting `≤ 1 wei drift` vs Anvil receipts across 30+ historical swaps per protocol is its own (tedious) PR — deliberately separated so this PR stays reviewable.
- **Curve 3-coin / tricrypto / metapool variants.** The 2-coin pinning matches the existing `get_amount_out` semantics; N-coin support requires the caller to pick `j` explicitly and is its own design.
- **Balancer composable / stable / weighted-N-token (N > 2).** Same 2-token pinning rationale.
- **Curve / Balancer event-driven cache refresh.** Bootstrap populates the cache; live refresh on Curve `TokenExchange` / Balancer `Swap` events needs new `PoolEvent` variants the event decoder doesn't emit yet. Until then the cache for those protocols is stale-but-correct between bootstrap and a restart.
- **Mempool decode pipeline integration.** The wrapper is callable but unused on this branch; the caller is `try_post_state_scan` in PR #118 (which currently routes V3/Curve/Balancer to `protocol_unsupported`). Plan: rebase #118 on this once it merges, swap the dispatch.

## Commit-by-commit

```
2a38f1c feat(pools,metrics): predict_post_state_with_fallback + EVM fallback metric
b945be4 feat(engine): populate PoolStateCache for Curve + Balancer pools
4fc2794 feat(engine): populate PoolStateCache for V3 pools
11eb605 feat(engine): populate PoolStateCache for V2/Sushi pools
961ae56 feat(pools,engine): PoolStateCache wired (empty) into AetherEngine
4328f1b feat(pools): BalancerPool::predict_post_state (weighted-2-token)
6e2b176 feat(pools): CurvePool::predict_post_state (2-coin stableswap)
550a263 feat(pools): V3 tick-crossing detection on predict_post_state
36160ec feat(pools): UniswapV3Pool::predict_post_state (single-tick)
```

Net diff: roughly +1,800 lines, mostly tests + math docs. Each commit is independently buildable + testable so reviewers can walk top-to-bottom or jump to a specific protocol family.

### DIFF
--- a/crates/grpc-server/src/engine.rs
+++ b/crates/grpc-server/src/engine.rs
@@ -89,6 +89,11 @@ pub struct PoolMetadata {
     pub pool_id: PoolId,
     pub protocol: ProtocolType,
     pub fee_bps: u32,
+    /// V3-only — tick spacing from `pools.toml`. `None` for non-V3
+    /// protocols (where the field is meaningless) and as a fallback
+    /// when the config entry omitted it. Required to construct a valid
+    /// `UniswapV3Pool` for the post-state cache.
+    pub tick_spacing: Option<i32>,
 }
 
 impl PoolMetadata {
@@ -429,6 +434,25 @@ impl AetherEngine {
         protocol: ProtocolType,
         fee_bps: u32,
     ) {
+        self.register_pool_with_tick_spacing(pool_addr, token0, token1, protocol, fee_bps, None)
+            .await
+    }
+
+    /// Same as [`Self::register_pool`] but accepts a tick_spacing hint
+    /// for V3-family pools. Required so the post-state cache can build
+    /// a valid `UniswapV3Pool` (which needs tick_spacing to know where
+    /// the active tick bucket ends). Non-V3 callers pass `None` and the
+    /// metadata stores it as `None` — the field is ignored at the
+    /// graph-edge update path which doesn't care about ticks.
+    pub async fn register_pool_with_tick_spacing(
+        &self,
+        pool_addr: Address,
+        token0: Address,
+        token1: Address,
+        protocol: ProtocolType,
+        fee_bps: u32,
+        tick_spacing: Option<i32>,
+    ) {
         let (t0_idx, t1_idx, num_tokens) = {
             let mut ti = (**self.token_index.load()).clone();
             let t0 = ti.get_or_insert(token0);
@@ -450,6 +474,7 @@ impl AetherEngine {
             pool_id,
             protocol,
             fee_bps,
+            tick_spacing,
         };
 
         {
@@ -609,7 +634,9 @@ impl AetherEngine {
                 continue;
             }
 
-            self.register_pool(pool_addr, token0, token1, protocol, entry.fee_bps)
+            self.register_pool_with_tick_spacing(
+                pool_addr, token0, token1, protocol, entry.fee_bps, entry.tick_spacing,
+            )
                 .await;
             loaded += 1;
 
@@ -660,7 +687,7 @@ impl AetherEngine {
         // Result type for each concurrent RPC fetch.
         enum ReserveResult {
             V2 { pool_addr: Address, meta: PoolMetadata, r0: U256, r1: U256 },
-            V3 { pool_addr: Address, meta: PoolMetadata, sqrt_price_x96: U256 },
+            V3 { pool_addr: Address, meta: PoolMetadata, sqrt_price_x96: U256, tick: i32 },
             Skipped,
         }
 
@@ -702,7 +729,20 @@ impl AetherEngine {
                         match provider.call(tx).await {
                             Ok(output) if output.len() >= 64 => {
                                 let sqrt_price_x96 = U256::from_be_slice(&output[0..32]);
-                                ReserveResult::V3 { pool_addr, meta, sqrt_price_x96 }
+                                // slot0 returns int24 tick, ABI-encoded as a
+                                // sign-extended 32-byte word at offset 32. Read
+                                // the high byte's MSB to detect a negative
+                                // value, then take the low 24 bits and apply
+                                // two's-complement if needed.
+                                let raw = &output[32..64];
+                                let mut tick_low24: i32 = ((raw[29] as i32) << 16)
+                                    | ((raw[30] as i32) << 8)
+                                    | (raw[31] as i32);
+                                if raw[0] != 0 {
+                                    // Top bytes set → negative; sign-extend.
+                                    tick_low24 -= 1 << 24;
+                                }
+                                ReserveResult::V3 { pool_addr, meta, sqrt_price_x96, tick: tick_low24 }
                             }
                             Ok(output) => {
                                 warn!(%pool_addr, len = output.len(), "slot0 output too short");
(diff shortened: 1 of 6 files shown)


nikhilbajaj31 (dismissed):
(superseded — see active review on this PR)

Pablosinyores (approved):
## Review: V3 / Curve / Balancer post-state predictors + state cache

### Verdict: APPROVE
**Blocking**: 0 CRITICAL · **Non-blocking**: 2 WARNING, 2 SUGGESTION

### What works
- All three predictors reuse the exact computation paths as the existing `get_amount_out` implementations — parity proven by tests.
- V3 tick-crossing detection uses proper Euclidean floor-division and handles negative ticks correctly.
- `PoolStateCache` design (`Arc<DashMap>` + per-entry `Arc<PoolState>`) gives readers snapshot isolation without blocking writers — sound for the hot path.
- New `aether_sim_evm_fallback_total` metric follows naming conventions (prefix, `_total` suffix, stable label set).
- 55 new tests cover predictors, cache, dispatch, and engine integration. Clippy clean, 427 tests passing, no secrets.

### Findings

**WARNING — V3 bootstrap seeds cache with `liquidity = 0`**
`crates/grpc-server/src/engine.rs` (V3 ReserveResult handler, around `v3.update_sqrt_price(sqrt_price_x96, 0u128, tick)`).
`slot0()` does not return liquidity, so `predict_post_state` returns `None` for every V3 pool until the first `V3Update` event. In a quiet market after restart, that window can be minutes. Suggest adding a `liquidity()` RPC call to the V3 bootstrap fanout — single word view, one extra call per V3 pool, closes the gap. The PR body acknowledges and defers this; acceptable for a foundation PR but should be the first follow-up.

**WARNING — Curve/Balancer cache is bootstrap-only**
`crates/grpc-server/src/engine.rs` `handle_pool_update`.
Curve `TokenExchange` and Balancer `Swap` events do not yet refresh the cache. Once #118 starts consuming this cache, stale Curve/Balancer state will silently produce incorrect post-state predictions that pass the `analytical = true` confidence check. Suggest at minimum a `warn!` log on first read of an entry not updated since bootstrap, or a tracking issue so this is not forgotten.

**SUGGESTION — `sqrt_price_x96_to_tick` doc scope**
`crates/pools/src/uniswap_v3.rs:28-43`.
The f64 path is sound for bucket-boundary detection but the on-chain implementation uses log-base-2 bit manipulation. If this function is ever used as a canonical tick oracle (e.g., for graph edge weights), the f64 path can accumulate error near MIN/MAX_TICK. Mark scope in the doc comment, e.g. `/// Used only for bucket-boundary detection, not as a canonical tick oracle`.

**SUGGESTION — New metric not in spec table**
`aether_sim_evm_fallback_total` is not listed in the `Key Observability Metrics` table in CLAUDE.md. Adding it (Counter, suggested alert threshold) ensures Grafana / AlertManager can be pre-configured. Not blocking since no dashboard wires it yet, but should land before the mempool pipeline goes live.

### Test coverage
| Area | Status |
|------|--------|
| V3 predict_post_state | Covered (8 tests) |
| Curve predict_post_state | Covered (7 tests) |
| Balancer predict_post_state | Covered (7 tests) |
| PoolStateCache types | Covered (3 tests) |
| Fallback dispatch | Covered (5 tests) |
| Engine cache population | Covered (5 tests) |
| Curve/Balancer bootstrap RPC end-to-end | Deferred to shadow run (acknowledged in PR body) |

### Security
- No secrets, private keys, JWTs, or `.env` files in the diff.
- No `unsafe`, no `.unwrap()` in production paths, no `panic!`.
- Balancer vault address is the canonical public deployment.
- No Solidity changes.

